PR #2, opened 2021-07-13
SpannerDaoInterface.java class

biancamacias:
this PR adds the `SpannerDaoInterface.java` class which will allow a separation of the Java client and JDBC implementations of the methods. i'm anticipating that `SpannerDao.java` class will be renamed to something along the lines of `SpannerDaoJavaImpl.java`, but if it is, that will be in a separate PR. 

the exceptions thrown in the interface are also still `SpannerException` and not a custom exception -- this will also be changed in another PR later on (when I write the class). suggestions for the interface description would be great -- are we using Javadoc? is that standard?

this PR (in summary) accomplishes the following:

- [x] creates a `SpannerDaoInterface.java` class with javadocs for each method
- [x] renames `SpannerDao` to `SpannerDaoImpl` in anticipation of the incoming JDBC implementation
- [x] binds `SpannerDaoImpl` to the interface in the `DatabaseModule` class
- [x] updates the `FinAppService` class to now inject the interface


src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,26 @@
+/*
```
haseebajmal, line 1: Please add the copyright header similar to other files.

haseebajmal, line 1: Please add a copyright header similar to other files.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,26 @@
...
+package com.google.finapp;
+
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.*;
+
+import java.math.BigDecimal;
+
+public interface SpannerDaoInterface {
+
+  void createCustomer(ByteArray customerId, String name, String address) throws SpannerException;
```
haseebajmal, line 14: Can we create a new exception type or are you planning to do that in a separate PR ?

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,26 @@
+/*
+description goes here...
+ */
+
+package com.google.finapp;
+
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.*;
```
haseebajmal, line 8: Can we import specific classes that are being used instead of *

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.math.BigDecimal;
+
+public interface SpannerDaoInterface {
```
haseebajmal, line 12: At some point, we'd want to add javadocs for these classes and methods. It could be a good practice if you want to take that up now.

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -22,7 +22,7 @@
 
 import java.math.BigDecimal;
 
-final class SpannerDao {
+final class SpannerDao implements com.google.finapp.SpannerDaoInterface {
```
haseebajmal, line 25: You don't need to fully qualify the class. Just implements SpannerDaoInterface should suffice.

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -31,7 +31,8 @@
     this.databaseClient = databaseClient;
   }
 
-  void createCustomer(ByteArray customerId, String name, String address) throws SpannerException {
+  @Override
+  public void createCustomer(ByteArray customerId, String name, String address) throws SpannerException {
```
haseebajmal, line 35: If the server code which uses this were in a separate package, we'd need these methods to be public. We'd want to do that in future I suppose for better code organization, but with the current state of packages, we don't need this to be public.

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -31,7 +31,8 @@
     this.databaseClient = databaseClient;
   }
 
-  void createCustomer(ByteArray customerId, String name, String address) throws SpannerException {
+  @Override
```
haseebajmal, line 34: Can we rename the class SpannerDao to convey that this implementation uses spanner's java client as opposed to jdbc.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,26 @@
...
+description goes here...
+ */
+
+package com.google.finapp;
+
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.*;
+
+import java.math.BigDecimal;
+
```
haseebajmal, line 11: A couple of more changes which need to be made:

- In the server code where we inject SpannerDao, we need to instead inject SpannerDaoInterface.
- In the database module we should bind SpannerDaoInterface to the implementation - https://www.baeldung.com/guice#2-guice-bindings

haseebajmal (changes requested):

src/main/java/com/google/finapp/SpannerDaoInterface.java
biancamacias: im going to do this in a separate PR

src/main/java/com/google/finapp/SpannerDao.java
biancamacias: i think this happened as a result of creating the interface (an intellij feature), so i'll resolve this

src/main/java/com/google/finapp/SpannerDaoInterface.java
biancamacias: do you mean in `FinAppService` to change from the `SpannerDao` to `SpannerDaoInterface`?

src/main/java/com/google/finapp/SpannerDao.java
biancamacias: if i dont specify that the methods are public, i get build failures. 
image

src/main/java/com/google/finapp/SpannerDaoJavaImpl.java
```diff
@@ -22,16 +22,17 @@
 
 import java.math.BigDecimal;
 
-final class SpannerDao {
+final class SpannerDaoJavaImpl implements SpannerDaoInterface {
```
c24t, line 25: I think `SpannerDaoImpl` (and later `SpannerDaoJDBCImpl`) would be fine too, up to you.

src/main/java/com/google/finapp/SpannerDaoJavaImpl.java
```diff
@@ -103,9 +107,10 @@ void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, BigDecim
             });
   }
 
-  void getAccountMetadata(ByteArray accountId) throws SpannerException {
-
-  }
+  // @Override
+  // public void getAccountMetadata(ByteArray accountId) throws SpannerException {
+  //
+  // }
```
c24t, line 113: Feel free to delete altogether until we're ready to implement it.

c24t (approved):
LGTM after addressing @haseebaj's comments

> suggestions for the interface description would be great -- are we using Javadoc?

I think javadoc is a great idea, we can track in a separate issue.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,74 @@
...
+   * @param accountStatus indicates unspecificed, active, or frozen Account status
+   * @param balance Account balance
+   * @throws SpannerException
+   */
+  void createAccount(
+      ByteArray accountId, AccountType accountType, AccountStatus accountStatus, BigDecimal balance)
+      throws SpannerException;
+
+  /**
+   * Inserts a new row to the Account table for a Customer in the database.
```
ngcgarcia, line 51: This function adds a row to the CustomerRole table, not the Account table.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,74 @@
...
+   * @param roleId unique role id
+   * @param roleName role name
+   * @throws SpannerException
+   */
+  void addAccountForCustomer(
+      ByteArray customerId, ByteArray accountId, ByteArray roleId, String roleName)
+      throws SpannerException;
+
+  /**
+   * Moves an amount from one unique account to another unique account for a Customer
```
ngcgarcia, line 64: Include that this modifies the Account table and adds rows to the TransactionHistory table (one transaction for the account losing the amount, another for the account gaining the amount).

ngcgarcia (changes requested):

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,75 @@
...
+   * in the database. Also modifies Account table and adds rows to the TransactionHistory table
```
ngcgarcia, line 65: This "also" is slightly misleading, since the act of modifying the Account table and TransactionHistory table *is* moving the amount, rather than an additional action. Maybe remove the "also" and instead change the line to be something like "... in the database by modifying the Account table and adding rows..."

ngcgarcia (changes requested):

ngcgarcia (approved):
LGTM after addressing Haseeb's comments.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * The SpannerDaoInterface declares the methods to be used by its separate implementations.
+ */
+public interface SpannerDaoInterface {
+
+  /**
+   * Inserts a new row to the Customer table in the database.
+   *
+   * @param customerId unique the Customer id
```
haseebajmal, line 30: If you don't have any interesting details to add here, you can omit these @params and @throws clauses.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -0,0 +1,75 @@
...
+   * Moves an amount from one unique account to another unique account for a Customer
+   * in the database by modifying the unique accounts in the Account table and adding rows
+   * to the TransactionHistory table in the database.
+   *
+   * @param fromAccountId unique account id where amount will be transferred from
+   * @param toAccountId unique account id where amount will be transferred to
+   * @param amount amount transferred from fromAccountId to toAccountId
+   * @throws SpannerException
+   */
+  void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId,
```
haseebajmal, line 73: Are you using google-java-format plugin ? This formatting seems odd. Can you double check and format the code ?

haseebajmal (approved):
LG. Thanks for addressing all the comments. 

Added a couple of minor ones. PTAL.

src/main/java/com/google/finapp/SpannerDaoInterface.java
biancamacias: i'm pretty sure i ran it correctly? i used `reformat code` under the `code` tab.

PR #3, opened 2021-07-13
Remove setChannelProvider call

c24t:
Remove some GKE-specific gRPC caller code

cc @haseebaj, @biancamacias, @ngcgarcia


biancamacias (approved):
lgtm!

haseebajmal (approved):
We should probably remove this code altogether, including the flags for project, instance, etc., and rely on the gcloud integration of client libraries. But I can do that in a subsequent change.

ngcgarcia (approved):
LGTM

PR #5, opened 2021-07-14
Fix quote character in README

ngcgarcia:
This PR fixes a typo in the README where a right double quotation mark was used instead of a straight double quotation mark.


c24t (approved):
🚀

biancamacias (approved):

PR #6, opened 2021-07-16
Implement SpannerDao with JDBC

ngcgarcia:
This PR creates an implementation of `SpannerDaoInterface` that uses JDBC instead of Cloud Spanner's Java client. The `SpannerDaoJDBCImpl` allows developers who use JDBC to see how Cloud Spanner can work with this familiar API.

As the code is now, the original Java client implementation will be used rather than this one. A future PR will provide a flag to switch between implementations, as described in the issue https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/8. With this PR, the JDBC implementation can be used by manually changing `DatabaseModule.configure` to bind `SpannerDaoJDBCImpl.class` instead of `SpannerDaoImpl.class`.

Accomplished in this PR:
- provide JDBC analogue for Java client implementation
- include appropriate dependency in `pom.xml` file


src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
-            });
+
+  void moveAccountBalance(
+      ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
+      throws SQLException {
+    try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
+      // begin transaction
+      connection.setAutoCommit(false);
+      try (
+          PreparedStatement readStatement = connection.prepareStatement(
```
ngcgarcia, line 119: It seems to be common practice to instantiate `Statement`s and `PreparedStatement`s right after `Connection`s. However, these statements are passed later to helper functions to populate the `?` variables, distancing the code that modifies the SQL commands from the commands themselves. I added comments that contain the SQL commands to improve readability, but another option could be to pass the `connection` rather than `PreparedStatement`s.

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
-  private ImmutableMap<ByteArray, BigDecimal> readAccountBalances(
-      ByteArray fromAccountId, ByteArray toAccountId, TransactionContext transaction) {
-    ResultSet resultSet =
-        transaction.read(
-            "Account",
-            KeySet.newBuilder().addKey(Key.of(fromAccountId)).addKey(Key.of(toAccountId)).build(),
-            ImmutableList.of("AccountId", "Balance"));
 
-    ImmutableMap.Builder<ByteArray, BigDecimal> accountBalancesBuilder = ImmutableMap.builder();
+  private BigDecimal[] readAccountBalances(
```
ngcgarcia, line 155: We could also use a `Pair` object, as @c24t suggested. Or `List` or something similar.

The reason the `Map` approach was abandoned was because we cannot use `Map.get` with `byte[]` objects, since equality must be done with `Arrays.equals` (so the `get` cannot find the correct key). Another way would be to convert back and forth between `ByteArray` and `byte[]` objects between functions so that we could use object equality and `Map.get` with `ByteArray`s

biancamacias (changes requested):
aside from the issues that youve pointed out in the description, here are a few things i notice with this PR

- we won't be able to merge this in because you've edited `SpannerDao` and not edited based on the fact that there is now an interface. you'll need to pull the changes that are now in the main branch and rename this file `SpannerDaoJdbcImpl.java` or something else similar
- we also won't be able to merge because of the exceptions, but i'm working on that and understand you did this to make sure it runs
- once you pull the changes from the main branch, you will need to bind your implementation with the interface in the `DatabaseModule.java` class

with your PR as it is now, you won't be able to merge until you address these changes. i also understand that some of them depend on me (the exceptions), but i anticipate that you'll be able to work on the interface and implementation fixes without being blocked. really solid PR though, nice use of the APIs.

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
-            Mutation.newInsertBuilder("Customer")
-                .set("CustomerId")
-                .to(customerId)
-                .set("Name")
-                .to(name)
-                .set("Address")
-                .to(address)
-                .build()));
+  void createCustomer(ByteArray customerId, String name, String address) throws SQLException {
+    try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
```
voulg, line 54: I'd expect the connection to be initialization once in the constructor and re-used for all classes

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
-  private final DatabaseClient databaseClient;
+  private final String projectId;
+  private final String databaseId;
+  private final String instanceId;
+  private final String connectionUrl;
 
   @Inject
-  SpannerDao(DatabaseClient databaseClient) {
-    this.databaseClient = databaseClient;
+  SpannerDao(@ArgsModule.SpannerProjectId String spannerProjectId,
```
voulg, line 37: I saw in the PR description that re-using the SpannerDao class is intentional and will change, but why not change it to SpannerDaoJDBC as part of this PR?

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
+          connection.prepareStatement(
+              "INSERT INTO Customer\n"
+                  + "(CustomerId, Name, Address)\n"
+                  + "VALUES\n"
+                  + "(?, ?, ?)")) {
+        ps.setBytes(1, customerId.toByteArray());
+        ps.setString(2, name);
+        ps.setString(3, address);
+        ps.executeUpdate();
+        System.out.printf("Customer created: %s\n", name);
```
voulg, line 65: Eventually we should use a java.util.logging.Logger for such information. Something like:

1. Create the logger in the class constructor:
private static final Logger logger = Logger.getLogger(TheClassName.class.getName());

2. Here you can do something like:
logger.log(Level.INFO, "My amazing message");

src/main/java/com/google/finapp/SpannerDao.java
```diff
@@ -16,133 +16,183 @@
...
+
+  void moveAccountBalance(
+      ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
+      throws SQLException {
+    try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
+      // begin transaction
+      connection.setAutoCommit(false);
+      try (
+          PreparedStatement readStatement = connection.prepareStatement(
+              "SELECT AccountId, Balance FROM Account WHERE (AccountId = ? or AccountId = ?)"
```
voulg, line 120: It would be good to check that this won't result one account to go negative and throw an appropriate error.

src/main/java/com/google/finapp/SpannerDao.java
voulg: I'd like to better understand the disadvantages of ByteArray, is this captured somewhere?

ngcgarcia:
> aside from the issues that youve pointed out in the description, here are a few things i notice with this PR
> 
> * we won't be able to merge this in because you've edited `SpannerDao` and not edited based on the fact that there is now an interface. you'll need to pull the changes that are now in the main branch and rename this file `SpannerDaoJdbcImpl.java` or something else similar
> * we also won't be able to merge because of the exceptions, but i'm working on that and understand you did this to make sure it runs
> * once you pull the changes from the main branch, you will need to bind your implementation with the interface in the `DatabaseModule.java` class
> 
> with your PR as it is now, you won't be able to merge until you address these changes. i also understand that some of them depend on me (the exceptions), but i anticipate that you'll be able to work on the interface and implementation fixes without being blocked. really solid PR though, nice use of the APIs.

Yes, that all makes sense. I didn't pull the changes for this first draft so as to let the reviewers see the direct comparison between the original java client implementation and the JDBC implementation, but the next draft will be updated to be on a new file.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: It will be changed in the next draft of this PR. I left the first draft with the same file name so as to let reviewers see the direct comparison between the Java client and JDBC implementations.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: Yes, this is a good idea. This should be handled in a future PR.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: Looking into this now. All the JDBC examples I saw followed a very similar format of establishing connections and statements at the start of functions similar to this, but this could just be because they were small sample snippets. 

Establishing the connection like this is also used in the [Cloud Spanner JDBC examples](https://github.com/GoogleCloudPlatform/java-docs-samples/tree/9defa78e095878d74b727f1e9fb6a74f06f0f12f/spanner/jdbc), which again can be just a result of showing simple individual function examples. 

But, yes, this makes sense. I'll reuse the connection.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: I'm not sure if this specific problem is documented anywhere. There's no "disadvantages" per say, but the problem is that `ByteArray` is an object defined in `com.google.cloud`, and generic Java `PreparedStatement`s don't support them. Thus, we have to convert to `byte[]` objects to use JDBC, and `byte[]` objects have to be compared with `Arrays.equals`, while `ByteArray` objects can use the basic `equals` method. This is what I'm assuming is causing the problem for trying to use `Map.get` methods using two different `byte[]` objects that contain the same information, but are different objects in memory.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: This is a good idea. This change needs to happen in both Java client and JDBC implementations, and may require some error handling and signature changes depending on how we want to pass up this invalid argument error. Probably pass up some error and handle it in `FinAppService` with something similar to this [exception handling](https://github.com/GoogleCloudPlatform/cloud-spanner-samples/blob/21040e93f623441cefc79ef6b661e9efae1db3b5/src/main/java/com/google/finapp/FinAppService.java#L61-L69).

This will be another big-ish change, so I'm thinking we put this in a separate PR. Thoughts?

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: Actually, is it best practice to hold the same connection? This would mean that the connection would be used for the entire life of the server, since a singular SpannerDao instance is used. If this is the preferred practice, I'll implement it @voulg

ngcgarcia: "The open-source JDBC driver handles pooling internally..." "These pool connections are re-used..." via https://cloud.google.com/spanner/docs/open-source-jdbc#do_i_need_to_provide_my_own_connection_pool (thanks @haseebajmal)

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -39,7 +40,7 @@ public void createCustomer(Customer customer, StreamObserver<Empty> responseObse
           UuidConverter.getBytesFromUuid(UUID.randomUUID()),
           customer.getName(),
           customer.getAddress());
-    } catch (SpannerException e) {
+    } catch (SQLException e) {
```
haseebajmal, line 43: I think you need to wait for Bianca's change ? Otherwise, this'll be a breaking change.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+      @ArgsModule.SpannerDatabaseId String spannerDatabaseId) {
+    this.projectId = spannerProjectId;
+    this.databaseId = spannerDatabaseId;
+    this.instanceId = spannerInstanceId;
+    // use this URL to connect to the emulator
+    this.connectionUrl = String.format(
+        "jdbc:cloudspanner://localhost:9010/projects/%s/instances/%s/databases/%s;usePlainText=true",
+        projectId, instanceId, databaseId);
+    // use this URL to connect to Cloud Spanner
+    // this.connectionUrl = String.format(
```
haseebajmal, line 48: I would avoid checking in commented code. You can instead perhaps 
1. check if System.getenv("SPANNER_EMULATOR_HOST") gives you localhost and use that, else connect to prod. 
2. Or explicitly take spannerHost/port as arguments.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+        results[1] = resultSet.getBigDecimal("Balance");
+      }
+    }
+    return results;
+  }
+
+
+  private void updateAccount(byte[] accountId, BigDecimal newBalance,
+      PreparedStatement preparedStatement) throws SQLException {
+    // preparedStatement: "UPDATE Account SET Balance = ? WHERE AccountId = ?"
```
haseebajmal, line 176: Why not construct the PreparedStatement here - Makes it cleaner/bug-free IMO and avoids the need for this comment.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    preparedStatement.setBigDecimal(1, newBalance);
+    preparedStatement.setBytes(2, accountId);
+    preparedStatement.addBatch();
+  }
+
+  private void insertTransaction(byte[] fromAccountId, byte[] toAccountId, BigDecimal amount,
+      PreparedStatement preparedStatement) throws SQLException {
+    /* preparedStatement: "INSERT INTO TransactionHistory (AccountId, Amount, IsCredit, EventTimestamp)"
```
haseebajmal, line 184: Similar comment as updateAccount.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+      } else {
+        results[1] = resultSet.getBigDecimal("Balance");
+      }
+    }
+    return results;
+  }
+
+
+  private void updateAccount(byte[] accountId, BigDecimal newBalance,
+      PreparedStatement preparedStatement) throws SQLException {
```
haseebajmal, line 175: This formatting doesn't seem to be from google-java-format. Can you please run that ?

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+            fromAccountIdArray, toAccountIdArray, amount,
+            insertTransactionStatement);
+        insertTransactionStatement.executeUpdate();
+        connection.commit();
+        System.out.printf("Balance of %s moved.\n", amount.toString());
+      }
+    }
+  }
+
+  void getAccountMetadata(ByteArray accountId) throws SpannerException {
```
haseebajmal, line 149: You can remove this for now.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    this.connectionUrl = String.format(
+        "jdbc:cloudspanner://localhost:9010/projects/%s/instances/%s/databases/%s;usePlainText=true",
+        projectId, instanceId, databaseId);
+    // use this URL to connect to Cloud Spanner
+    // this.connectionUrl = String.format(
+    //     "jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s",
+    //     projectId, instanceId, databaseId);
+  }
+
+  public void createCustomer(ByteArray customerId, String name, String address) throws SQLException {
```
haseebajmal, line 53: You need to add @Override for all the public methods.

This as is, wouldn't compile IMO.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.finapp;
+
+import com.google.cloud.ByteArray;
+import com.google.cloud.spanner.*;
```
haseebajmal, line 18: Can we directly import required classes instead of *

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+import com.google.inject.Inject;
+
+import java.math.BigDecimal;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.Arrays;
+
+final class SpannerDaoJDBCImpl implements SpannerDaoInterface {
```
haseebajmal, line 29: Do we plan on adding a flag and switching the implementations based on that in https://github.com/GoogleCloudPlatform/cloud-spanner-samples/blob/main/src/main/java/com/google/finapp/DatabaseModule.java

Also please test this implementation out locally.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+          connection.prepareStatement(
+              "INSERT INTO Customer\n"
+                  + "(CustomerId, Name, Address)\n"
+                  + "VALUES\n"
+                  + "(?, ?, ?)")) {
+        ps.setBytes(1, customerId.toByteArray());
+        ps.setString(2, name);
+        ps.setString(3, address);
+        ps.executeUpdate();
+        System.out.printf("Customer created: %s\n", name);
```
haseebajmal, line 65: +1 to Panos's comment on using logger instead. I'd prefer it if we do it in this PR. 

For some reason, I'm not able to reply on that conversation.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+
+  void getAccountMetadata(ByteArray accountId) throws SpannerException {
+
+  }
+
+
+  private BigDecimal[] readAccountBalances(
+      byte[] fromAccountId, byte[] toAccountId, PreparedStatement preparedStatement)
+      throws SQLException {
+    // preparedStatement: "SELECT AccountId, Balance FROM Account WHERE (AccountId = ? or AccountId = ?)"
```
haseebajmal, line 157: Similar comment as others. Please avoid this.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+    }
+  }
+
+
+  public void moveAccountBalance(
+      ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
+      throws SQLException {
+    try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
+      // begin transaction
```
haseebajmal, line 116: Is this comment necessary ? There's only one transaction which spans the entire method.

haseebajmal (changes requested):

src/main/java/com/google/finapp/FinAppService.java
ngcgarcia: Yes, this PR will be edited and merged after Bianca's change.

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/11

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/12

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
ngcgarcia: The reason I added this comment is to explain that `connection.setAutoCommit(false)` is the action that sets up the transaction, since you turn off auto-commit in order to make the following `execute`s part of a transaction. However, I guess the target audience can probably understand that turning off auto-commit and having the `commit` at the end makes it a transaction.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
ngcgarcia: I'm going to address logging in a PR immediately following this. I'll have to update both `SpannerDaoImpl` and `SpannerDaoJDBCImpl`, and I want to keep this PR to only modifying/creating `SpannerDaoJDBCImpl`. I'd prefer to setup logging in a single PR. I also want to improve the statements we use for logging. See issue https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/11

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
ngcgarcia: Will be handled with https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/8.

This implementation has been tested (at least with the given RPCs) locally.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
ngcgarcia: This compiles without the `@override`. I believe that the annotation is needed if there is a parent implementation of the method that needs to be overriden. Since there is no parent implementation, it should be fine without `@override`

src/main/java/com/google/finapp/SpannerDao.java
ngcgarcia: For longevity, here is the blurb I originally had in the PR description:

> `PreparedStatement` does not support the `ByteArray` object that is in the signature for most methods. Thus, we must convert these `ByteArray` objects to `byte[]` objects. This complicates things in `moveAccountBalance` and its helper functions, where both reading and writing to the database are required. These `byte[]` objects cannot be compared with basic object equality, instead requiring `Arrays.equals`. Options include continuously switching between `byte[]` and `ByteArray` to support interacting with the database and object equality, or--as I have implemented--sticking with `byte[]` objects and changing the structure of some helper functions.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,191 @@
...
+              "jdbc:cloudspanner:/projects/%s/instances/%s/databases/%s",
+              projectId, instanceId, databaseId);
+    }
+  }
+
+  public void createCustomer(ByteArray customerId, String name, String address)
+      throws SpannerDaoException {
+    try {
+      try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
+        try (PreparedStatement ps =
```
voulg, line 61: A few questions:
1. Why do you need the outer-most try?
2. Also have you considered defining multiple resources with a single try-with-resources, e.g: try this:
```
try (Connection connection = DriverManager.getConnection(this.connectionUrl);
       PreparedStatement ps = connection....blahblah...) {
```
same comment applies for other methods.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,191 @@
...
+
+  public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
+      throws SpannerDaoException {
+    try {
+      try (Connection connection = DriverManager.getConnection(this.connectionUrl)) {
+        connection.setAutoCommit(false);
+        byte[] fromAccountIdArray = fromAccountId.toByteArray();
+        byte[] toAccountIdArray = toAccountId.toByteArray();
+        BigDecimal[] accountBalances =
+            readAccountBalances(fromAccountIdArray, toAccountIdArray, connection);
```
voulg, line 132: Do you think you might use readAccountBalances anywhere else? If not, have you considered inlining the logic here?

This would allow you to define separate:
BigDecimal sourceAmount;
BigDecimal destAmount;
and avoid having to deal with BigDecimal[] and indexes that are not as readable.

What do you think?

ngcgarcia: Good idea. I mainly did this separate function to abstract away database interactions into separate helper functions (like `updateAccount` and `insertTransaction`) to make the logic clearer, rather than for the sake of reuse. However, at this point, this helper function is being more confusing than helpful, so I'll move it inline.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,176 @@
...
+            connection.prepareStatement(
+                "SELECT AccountId, Balance FROM Account WHERE (AccountId = ? or AccountId = ?)")) {
+      connection.setAutoCommit(false);
+      byte[] fromAccountIdArray = fromAccountId.toByteArray();
+      byte[] toAccountIdArray = toAccountId.toByteArray();
+      readStatement.setBytes(1, fromAccountIdArray);
+      readStatement.setBytes(2, toAccountIdArray);
+      java.sql.ResultSet resultSet = readStatement.executeQuery();
+      BigDecimal sourceAmount = new BigDecimal(Long.MIN_VALUE);
+      BigDecimal destAmount = new BigDecimal(Long.MIN_VALUE);
```
ngcgarcia, line 128: I had to initialize the `BigDecimal`s here to resolve the compilation errors (since the following code doesn't guarantee that `sourceAmount` and `destAmount` will be assigned). See later comment.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,176 @@
...
+      while (resultSet.next()) {
+        byte[] currentId = resultSet.getBytes("AccountId");
+        if (Arrays.equals(currentId, fromAccountIdArray)) {
+          sourceAmount = resultSet.getBigDecimal("Balance");
+        } else {
+          destAmount = resultSet.getBigDecimal("Balance");
+        }
+      }
+      if (sourceAmount.min(destAmount).equals(new BigDecimal(Long.MIN_VALUE))) {
+        throw new IllegalArgumentException();
```
ngcgarcia, line 138: Since I initialized the values to be `Big Decimal(Long.MIN_VALUE)`, I figured I could utilize that here to see if either the fromAccount or toAccount weren't found in the database. We haven't formally established this, but we could have an invariant where there are no negative balances in the database. With that invariant, we could set the `BigDecimal` value to -1 or something similar to simplify things. Thoughts @voulg ?

ngcgarcia: fyi `(new BigDecimal(Long.MIN_VALUE)).equals(new BigDecimal(Long.MIN_VALUE))` --> `true`

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,176 @@
...
+  private final String connectionUrl;
+
+  @Inject
+  SpannerDaoJDBCImpl(
+      @ArgsModule.SpannerProjectId String spannerProjectId,
+      @ArgsModule.SpannerInstanceId String spannerInstanceId,
+      @ArgsModule.SpannerDatabaseId String spannerDatabaseId) {
+    this.projectId = spannerProjectId;
+    this.databaseId = spannerDatabaseId;
+    this.instanceId = spannerInstanceId;
```
voulg, line 40: It doesn't seem you ever use these three afterwards. Should we remove them here and keep only connectionUrl as state?

voulg (approved):

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -16,7 +16,6 @@
 
 import com.google.inject.Inject;
 import com.google.protobuf.Empty;
-import com.google.finapp.SpannerDaoException;
```
biancamacias, line 19: how come you deleted this from this file?

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -0,0 +1,170 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.finapp;
+
+import com.google.cloud.ByteArray;
+import com.google.inject.Inject;
```
biancamacias, line 18:
```suggestion
import com.google.cloud.ByteArray;
import com.google.finapp.SpannerDaoException;
import com.google.inject.Inject;
```

biancamacias (changes requested):
when i finish #15, you should include in this PR to also update DatabaseModule to `return new SpannerDaoJDBCImpl(client);` in the `provideSpannerDao` method

src/main/java/com/google/finapp/FinAppService.java
ngcgarcia: The code compiled without it (and IntelliJ highlighted it as an unnecessary import). I believe since this exception is created in the same package, it doesn't need to be imported.

biancamacias (approved):

ngcgarcia:
> when i finish #15, you should include in this PR to also update DatabaseModule to `return new SpannerDaoJDBCImpl(client);` in the `provideSpannerDao` method

will be handled in #15

c24t (approved):

PR #15, opened 2021-07-20
Adding SpannerDaoImpl flag

biancamacias:
this PR addresses the issue of not being able to specify which implementation to use by adding a flag that can be set using the command line to switch between the `SpannerDaoImpl` and `SpannerDaoJdbcImpl`.

this PR does the following:

- [x] updates `ArgsModule` by adding a new flag `SpannerUseJdbc` which takes a boolean value (and by default uses `SpannerDaoImpl`)
- [x] updates `DatabaseModule` to create a new client based on the flag -- creates new classes without using Guice
- [x] creates dummy `SpannerDaoJdbcImpl` class


src/main/java/com/google/finapp/SpannerDaoJdbcImpl.java
```diff
@@ -0,0 +1,91 @@
+// Copyright 2021 Google LLC
```
voulg, line 1: I'd leave this file out for now, we can wait for Nick's PR.

Instead of creating the jdbc object you can just throw an exception that it's not implemented and we can add it later what do you think?

src/main/java/com/google/finapp/ArgsModule.java
```diff
@@ -91,6 +95,9 @@ String provideSpannerDatabaseId() {
 
     @Parameter(names = {"--spanner_database_id"})
     String spannerDatabaseId;
+
+    @Parameter(names = {"--spanner_dao_impl"})
```
voulg, line 99: Can you file an issue to add description and required in these annotations? In particular here it would be good to document what are the available choices.

src/main/java/com/google/finapp/DatabaseModule.java
```diff
@@ -28,21 +28,25 @@
...
+      @ArgsModule.SpannerDatabaseId String spannerDatabaseId,
+      @ArgsModule.SpannerDaoImpl String spannerDaoImpl) {
     SpannerOptions spannerOptions = SpannerOptions.getDefaultInstance();
     Spanner spanner = spannerOptions.toBuilder().build().getService();
-    return spanner.getDatabaseClient(
+    DatabaseClient client = spanner.getDatabaseClient(
         DatabaseId.of(spannerProjectId, spannerInstanceId, spannerDatabaseId));
+    if (spannerDaoImpl == "jdbc") {
+      return new SpannerDaoJdbcImpl(client);
+    } else {
```
voulg, line 48: It would be preferable behavior to check for some other string e.g: "client_lib".

Here's one reason, assume the customer mis-spelled, or capitalized, e.g: JDBC, the result would be to run with client library, but it would be preferable to throw an exception with the supported flag values.

src/main/java/com/google/finapp/SpannerDaoJdbcImpl.java
biancamacias: so something like...
`if (SpannerDaoImpl == "jdbc") { throw [some exception]; }`?

src/main/java/com/google/finapp/ArgsModule.java
```diff
@@ -91,6 +95,9 @@ String provideSpannerDatabaseId() {
...
+    String spannerDaoImpl;
```
ngcgarcia, line 100: Since we only have two options, I think we should use a boolean value here. [See this documentation](https://jcommander.org/#_boolean). With this, we can rename the parameter to something like `useJDBC` and have the default be set to `false`. Thus, no actual arguments are needed (just the flag by itself), and we don't have to check for any strings like "jdbc" or "client".

src/main/java/com/google/finapp/DatabaseModule.java
ngcgarcia: Look at my other comment about using a boolean instead. That's my opinion, at least.

c24t (commented):
+1 to using a bool-valued flag instead of passing "jdbc" by name, otherwise LGTM.

src/main/java/com/google/finapp/ArgsModule.java
```diff
@@ -72,6 +72,10 @@ String provideSpannerDatabaseId() {
     return args.spannerDatabaseId;
   }
 
+  @Provides
+  @SpannerUseJdbc
+  boolean provideSpannerUseJdbc() { return args.spannerUseJdbc; }
```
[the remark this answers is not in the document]
biancamacias: i think we i ran the formatter something weird happened, ill try again );

voulg (approved):
LGTM, but let's wait for at least one more approval...

biancamacias:
currently getting an error like this after changing it back to a boolean, so im a bit confused -- this was fixed last time by adding the arity annotation, but im getting it again...
```
$ java -jar target/finapp-1.0-SNAPSHOT-jar-with-dependencies.jar --spanner_project_id=test-project 
--spanner_instance_id=test-instance --spanner_database_id=test-database --spanner_use_jdbc=true
Exception in thread "main" com.beust.jcommander.ParameterException: Was passed main parameter 'true' but no main parameter was defined in your arg class
	at com.beust.jcommander.JCommander.initMainParameterValue(JCommander.java:954)
	at com.beust.jcommander.JCommander.parseValues(JCommander.java:755)
	at com.beust.jcommander.JCommander.parse(JCommander.java:356)
	at com.beust.jcommander.JCommander.parse(JCommander.java:335)
	at com.google.finapp.ArgsModule.<init>(ArgsModule.java:33)
	at com.google.finapp.ServerMain.main(ServerMain.java:24)
```

src/main/java/com/google/finapp/ArgsModule.java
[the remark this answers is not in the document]
ngcgarcia: Bump on this.

src/main/java/com/google/finapp/DatabaseModule.java
```diff
@@ -28,21 +28,25 @@
...
     SpannerOptions spannerOptions = SpannerOptions.getDefaultInstance();
     Spanner spanner = spannerOptions.toBuilder().build().getService();
-    return spanner.getDatabaseClient(
+    DatabaseClient client = spanner.getDatabaseClient(
         DatabaseId.of(spannerProjectId, spannerInstanceId, spannerDatabaseId));
+    if (spannerUseJdbc) {
+      return new SpannerDaoJDBCImpl(spannerProjectId, spannerInstanceId, spannerDatabaseId);
+    } else {
+      return new SpannerDaoImpl(client);
+    }
```
ngcgarcia, line 50: If `spannerUseJdbc`, then we don't need to initialize `spannerOptions`, `spanner`, or `client`. Maybe move those all into the `else` clause (to both make it clear that that is part of the other implementation and to not waste any computation).

ngcgarcia (changes requested):

src/main/java/com/google/finapp/ArgsModule.java
```diff
@@ -91,6 +98,9 @@ String provideSpannerDatabaseId() {
 
     @Parameter(names = {"--spanner_database_id"})
     String spannerDatabaseId;
+
+    @Parameter(names = {"--spanner_use_jdbc"}, arity = 0)
```
voulg, line 102: Optional: If we want `arity = 0`, we can safely remove it,  this is the default.

voulg (approved):

haseebajmal (approved):
Should we update the README to mention this flag ?

ngcgarcia (approved):

biancamacias:
> Should we update the README to mention this flag ?

@haseebajmal i think so. what would that look like if i did? currently the `README` doesnt reflect how i'm running the application locally but im not sure how we expect users to run it.

c24t (approved):

Issue #12, opened 2021-07-19
Check for invalid inputs in moveAccountBalance


ngcgarcia:
## Issue

In `moveAccountBalance` of `SpannerDaoImpl` (and soon `SpannerDaoJDBCImpl`), we currently do not check for invalid inputs for `amount`, meaning we allow for a greater `amount` to be transferred from an account than what is in the account, making the account balance go negative. We should not allow account balances to go negative.

## Next Steps

- [x] Define output for invalid `amount` input
- [x] Update `SpannerDaoImpl` 
- [x] Update `SpannerDaoJDBCImpl` (when merged)

ngcgarcia:
Closed with #17

PR #17, opened 2021-07-21
Ensure non-negative account balances

ngcgarcia:
Addresses #12 

This PR ensures that accounts will not have negative balances (at least, with the current supported functions) by checking for: 
- negative inputs to `createAccount` 
- transfer amounts in `moveAccountBalance` that would make an account balance negative

On a separate, related note, this PR also checks for negative inputs to `moveAccountBalance` to enforce that the `IsCredit` bool in `TransactionHistory` is correct (a non-negative amount is being taken from `fromAccount` and given to `toAccount`).

When running code reformatting on this PR, several imports were changed (to improve code style). This includes the removal of unnecessary imports and blank lines and importing specific classes from Cloud Spanner instead of all. 

Accomplished in this PR:
- checks for negative inputs to `createAccount`
- checks for transfer amounts in `moveAccountBalance`
- checks for negative inputs to `moveAccountBalance`
- updates documentation to disallow invalid inputs
- clean up imports in related files

### DIFF
--- a/src/main/java/com/google/finapp/SpannerDaoImpl.java
+++ b/src/main/java/com/google/finapp/SpannerDaoImpl.java
@@ -61,7 +61,9 @@ final class SpannerDaoImpl implements SpannerDaoInterface {
       ByteArray accountId, AccountType accountType, AccountStatus accountStatus, BigDecimal balance)
       throws SpannerDaoException {
     if (balance.signum() == -1) {
-      throw new IllegalArgumentException("Account balance cannot be negative");
+      throw new IllegalArgumentException(String.format(
+          "Account balance cannot be negative. accountId: %s, balance: %s", accountId.toString(),
+          balance.toString()));
     }
     try {
       databaseClient.write(
@@ -109,7 +111,8 @@ final class SpannerDaoImpl implements SpannerDaoInterface {
   public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
       throws SpannerDaoException {
     if (amount.signum() == -1) {
-      throw new IllegalArgumentException("Amount cannot be negative");
+      throw new IllegalArgumentException(String.format(
+          "Amount transferred cannot be negative. amount: %s", amount.toString()));
     }
     try {
       databaseClient
@@ -123,7 +126,9 @@ final class SpannerDaoImpl implements SpannerDaoInterface {
                 BigDecimal newSourceAmount = accountBalances.get(fromAccountId).subtract(amount);
 
                 if (newSourceAmount.signum() == -1) {
-                  throw new IllegalArgumentException("Account balance cannot be negative");
+                  throw new IllegalArgumentException(String.format(
+                      "Account balance cannot be negative. original account balance: %s, amount transferred: %s",
+                      accountBalances.get(fromAccountId).toString(), amount.toString()));
                 }
 
                 transaction.buffer(
--- a/src/main/java/com/google/finapp/SpannerDaoImpl.java
+++ b/src/main/java/com/google/finapp/SpannerDaoImpl.java
@@ -60,7 +60,7 @@ final class SpannerDaoImpl implements SpannerDaoInterface {
   public void createAccount(
       ByteArray accountId, AccountType accountType, AccountStatus accountStatus, BigDecimal balance)
       throws SpannerDaoException {
-    if (balance.compareTo(BigDecimal.ZERO) < 0) {
+    if (balance.signum() == -1) {
       throw new IllegalArgumentException("Account balance cannot be negative");
     }
     try {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -55,6 +60,9 @@ public void createCustomer(ByteArray customerId, String name, String address)
   public void createAccount(
       ByteArray accountId, AccountType accountType, AccountStatus accountStatus, BigDecimal balance)
       throws SpannerDaoException {
+    if (balance.signum() == -1) {
+      throw new IllegalArgumentException("Account balance cannot be negative");
+    }
```
ngcgarcia, line 65: Any of these `if` `throw` constructions throughout the PR can also be assertions. I chose this method so I could specifically use `IllegalArgumentException` to indicate that something the user input was incorrect, but if assertions are the preferred method, this can be switched for any or all of these instances.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -129,7 +135,10 @@ public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, B
         }
       }
       if (sourceAmount == null || destAmount == null) {
-        throw new IllegalArgumentException();
+        throw new IllegalArgumentException("Account not found");
```
ngcgarcia, line 138: I figured I should add a small string explanation here.

biancamacias, line 138: is there a way we could specify which account is not found? maybe by its account id or something? not necessary but could be helpful when testing.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -129,7 +135,10 @@ public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, B
...
+      }
+      if (sourceAmount.subtract(amount).signum() == -1) {
```
c24t, line 140: Make a `newSourceAmount` instead of calling `subtract` twice?

src/main/java/com/google/finapp/SpannerDaoImpl.java
c24t: `IllegalArgumentException` looks right to me.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -53,7 +53,8 @@ void addAccountForCustomer(
    *
    * @param fromAccountId unique account id where amount will be transferred from
    * @param toAccountId unique account id where amount will be transferred to
-   * @param amount amount transferred from fromAccountId to toAccountId
+   * @param amount amount transferred from fromAccountId to toAccountId, must be less than or equal
+   * to fromAccountId's account balance, must be non-negative
```
c24t, line 57: Good call documenting the preconditions.

c24t (approved):

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -32,8 +30,10 @@
   /**
    * Inserts a new row to the Account table in the database.
    *
+   * @param accountId unique id of account
```
biancamacias, line 33: some of these javadocs might be unnecessary? `accountId` and `balance` are self-explantory. just a suggestion to simplify the javadocs.

src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -55,6 +60,9 @@ public void createCustomer(ByteArray customerId, String name, String address)
   public void createAccount(
       ByteArray accountId, AccountType accountType, AccountStatus accountStatus, BigDecimal balance)
       throws SpannerDaoException {
+    if (balance.signum() == -1) {
+      throw new IllegalArgumentException("Account balance cannot be negative");
```
voulg, line 64: Optional: You could also put the negative balance that caused the issue in the exception string, this could help with debugging in a real application. Similar comment for all other exception strings.

src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -109,6 +120,10 @@ public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, B
                 ImmutableMap<ByteArray, BigDecimal> accountBalances =
                     readAccountBalances(fromAccountId, toAccountId, transaction);
 
+                if (accountBalances.get(fromAccountId).subtract(amount).signum() == -1) {
+                  throw new IllegalArgumentException("Account balance cannot be negative");
```
voulg, line 124: Similar comment, here it would be good to print both the starting amount and the amount that we tried to move.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
voulg: +1

src/main/java/com/google/finapp/SpannerDaoInterface.java
ngcgarcia: `accountId` can be removed, but `balance` description contains the non-negative precondition

biancamacias (approved):

voulg (approved):

PR #18, opened 2021-07-26
[on hold] Updating README content

### LINKED ISSUE #14: Update README
## Issue
with the new flag that switches between implementations and new command line instructions, we need to update the `README`.

- [ ] add documentation about how to switch the flag
- [ ] update commands about how to run the emulator
- [ ] specify separate terminal windows to run the emulator, grpc service, and to call the RPCs
- [ ] update commands about running the server

biancamacias:
this PR adds more content to the `README.md` file so that users are better able to run the application. There is also information on how to run either the java client or JDBC implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,14 @@ This repository contains a sample banking application built using Cloud Spanner.
 
 NOTE: Requires gcloud, mvn, grpc_cli installed.
 
-1. Create a database locally using cloud-spanner-emulator and export spanner host
+1. Create a database locally using [cloud-spanner-emulator](https://cloud.google.com/spanner/docs/emulator) and export spanner host
 for client libraries to work.
 
     ```
     $ gcloud emulators spanner start
+   ```
+   In a **separate terminal window**, if you are using the emulator **for the first time**, run:
+   ```
     $ gcloud config configurations create emulator
     $ gcloud config set auth/disable_credentials true
     $ gcloud config set project your-project-id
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,17 +25,42 @@ for client libraries to work.
...
+    $ gcloud config set auth/disable_credentials true
+    $ gcloud config set project your-project-id
+    $ gcloud config set api_endpoint_overrides/spanner http://localhost:9020/
+    $ gcloud spanner instances create test-instance \
+        --config=emulator-config --description="Test Instance" --nodes=1
+    $ gcloud config set spanner/instance test-instance
+    $ gcloud spanner databases create test-database \
+        --ddl-file src/main/java/com/google/finapp/schema.sdl
+    $ export SPANNER_EMULATOR_HOST="localhost:9010"
+    ```
```
ngcgarcia, line 40: A lot of this is repeated--if I'm correct, everything except the first line. Maybe instead say something like:

 In a **separate terminal window**, if you are using the emulator **for the first time**, run:
`    $ gcloud config configurations create emulator`

If you are **not** using the emulator for the first time, run:
`    $ gcloud config configurations activate emulator`

Then run: 
_everything else_

README.md
```diff
@@ -6,11 +6,14 @@ This repository contains a sample banking application built using Cloud Spanner.
 
 NOTE: Requires gcloud, mvn, grpc_cli installed.
 
-1. Create a database locally using cloud-spanner-emulator and export spanner host
+1. Create a database locally using [cloud-spanner-emulator](https://cloud.google.com/spanner/docs/emulator) and export spanner host
 for client libraries to work.
 
     ```
     $ gcloud emulators spanner start
```
ngcgarcia, line 13: This looks mildly misleading. At least the way I first read this made it seem like this singular line is what did the "Create a database locally using cloud-spanner-emulator and export spanner host for client libraries to work" since this line is now separated from the rest of the code that was originally under this statement. Maybe it would seem a bit clearer if you did something like:

Create a database locally using cloud-spanner-emulator and export spanner host for client libraries to work.
Begin by starting an emulator in one terminal window.

ngcgarcia (changes requested):
Great idea to update the README. I would say that you should also talk to @haseebajmal to make sure that the way that we've been running the application is the intended way for end users to.

PR #20, opened 2021-07-30
addAccountForCustomer rpc

biancamacias:
this PR creates the rpc for `addAccountForCustomer` method:

- [x] adds rpc to `service.proto`
- [x] adds message request to `service.proto`
- [x] adds new method to `FinAppService` that calls `spannerDao.addAccountForCustomer` with the appropriate parameters

this PR does not include (yet):
1. tests
2. change the RPCs to return anything

### DIFF
--- a/src/main/proto/service.proto
+++ b/src/main/proto/service.proto
@@ -55,8 +55,8 @@ message Account {
 }
 
 message AddAccountForCustomerRequest {
-  optional bytes customerId = 1;
-  optional bytes accountId = 2;
-  optional bytes roleId = 3;
-  optional string roleName = 4;
+  optional bytes customer_id = 1;
+  optional bytes account_id = 2;
+  optional bytes role_id = 3;
+  optional string role_name = 4;
 }
--- a/src/main/proto/service.proto
+++ b/src/main/proto/service.proto
@@ -28,7 +28,7 @@ service FinApp {
 
   rpc CreateAccount(Account) returns (.google.protobuf.Empty) {}
 
-  rpc addAccountForCustomer(AddAccountForCustomerRequest) returns (.google.protobuf.Empty) {}
+  rpc addAccountForCustomer(CustomerRole) returns (.google.protobuf.Empty) {}
 }
 
 message Customer {
--- a/src/main/proto/service.proto
+++ b/src/main/proto/service.proto
@@ -28,7 +28,7 @@ service FinApp {
 
   rpc CreateAccount(Account) returns (.google.protobuf.Empty) {}
 
-  rpc addAccountForCustomer(CustomerRole) returns (.google.protobuf.Empty) {}
+  rpc AddAccountForCustomer(CustomerRole) returns (.google.protobuf.Empty) {}
 }
 
 message Customer {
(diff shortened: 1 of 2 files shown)


src/main/proto/service.proto
```diff
@@ -50,4 +52,11 @@ message Account {
...
   optional Status status = 3;
-}
\ No newline at end of file
+}
+
+message AddAccountForCustomerRequest {
+  optional bytes customerId = 1;
+  optional bytes accountId = 2;
+  optional bytes roleId = 3;
+  optional string roleName = 4;
```
voulg, line 61: https://developers.google.com/protocol-buffers/docs/style#message_and_field_names
```
Use underscore_separated_names for field names (including oneof field and extension names) – for example, song_name.
```

src/main/proto/service.proto
```diff
@@ -50,4 +52,11 @@ message Account {
...
   }
   optional Status status = 3;
-}
\ No newline at end of file
+}
+
+message AddAccountForCustomerRequest {
+  optional bytes customerId = 1;
+  optional bytes accountId = 2;
+  optional bytes roleId = 3;
```
voulg, line 60: The role id is the unique identifier for the new role. I'd prefer if this is generated by the backend and returned in the response. (We should also return customer_id / account_id in the CreateX methods, but this can be done in a separate PR)

src/main/proto/service.proto
```diff
@@ -27,6 +27,8 @@ service FinApp {
   rpc CreateCustomer(Customer) returns (.google.protobuf.Empty) {}
 
   rpc CreateAccount(Account) returns (.google.protobuf.Empty) {}
+
+  rpc addAccountForCustomer(AddAccountForCustomerRequest) returns (.google.protobuf.Empty) {}
```
voulg, line 31: https://developers.google.com/protocol-buffers/docs/style#message_and_field_names
```
Use CamelCase (with an initial capital) for message names – for example, SongServerRequest.
```

voulg: In retrospect, `CreateCustomerRoleForAccount` might be a better name. What do you think?

ngcgarcia: +1

ngcgarcia (approved):
Follow up on @voulg's comments. Other than that, LGTM!

src/main/proto/service.proto
biancamacias: i think ill do this in a separate PR because it means changing the signature in a lot of files, which might crowd this PR.

src/main/proto/service.proto
biancamacias: i changed the name to mimic the other requests, but lmk if i should change it (and maybe even the other requests to say explicitly `...Request`

voulg: Perhaps file a new issue to do the rename to CreateCustomerRoleForAccount and return UUID as a separate PR.

However as part of this PR we should follow CamelCase for message names, i.e:
AddAccountForCustomer

voulg (approved):
Two minor requests:
- Rename addAccountForCustomer to AddAccountForCustomer as part of this PR
- File an issue to rename it to CreateCustomerRoleForAccount and return a UUID with a future PR

Apart from these LGTM

src/main/proto/service.proto
biancamacias: ok i filed this in #25

PR #21, opened 2021-08-02
Establish test suite

ngcgarcia:
This PR establishes a test suite for the financial app. This PR does not create tests for all current methods; it only sets up a framework with one example. 

In its current state, the tests assume that the user has an emulator running in a separate terminal and has set up the string variables for project, instance, and database IDs.

This PR is also just a start to setting up tests. More work will be done on improving the tests, but this provides a baseline. 

Accomplished in this PR:
- set up test file
- create first method IT 
- add plugin to enable `mvn test`
- update README to give instructions for running tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,3 +36,10 @@ for client libraries to work.
     $ grpc_cli call localhost:8080 CreateCustomer \
         "name: 'google' address: 'amphitheatre pkwy'"
     ```
+
+## How to run the application tests
+
+1. Set up the emulator as described in #1 above.
+2. Replace the variables in `FinAppTests.setUpTests` according to your chosen instance, project, and database IDs.
+> The default variables should work for instance and database. The project variable must be changed based on `your-project-id` from `gcloud config set project your-project-id`. If this exact command was used, use `"your-project-id"`.
+3. Run `mvn test`.
--- a/README.md
+++ b/README.md
@@ -40,6 +40,4 @@ for client libraries to work.
 ## How to run the application tests
 
 1. Set up the emulator as described in #1 above.
-2. Replace the variables in `FinAppTests.setUpTests` according to your chosen instance, project, and database IDs.
-> The default variables should work for instance and database. The project variable must be changed based on `your-project-id` from `gcloud config set project your-project-id`. If this exact command was used, use `"your-project-id"`.
-3. Run `mvn test`.
+2. Run `mvn test`.
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ for client libraries to work.
     $ gcloud emulators spanner start
     $ gcloud config configurations create emulator
     $ gcloud config set auth/disable_credentials true
-    $ gcloud config set project your-project-id
+    $ gcloud config set project test-project
     $ gcloud config set api_endpoint_overrides/spanner http://localhost:9020/
     $ gcloud spanner instances create test-instance \
         --config=emulator-config --description="Test Instance" --nodes=1
--- a/src/test/java/com/google/finapp/FinAppIT.java
+++ b/src/test/java/com/google/finapp/FinAppIT.java
@@ -32,7 +32,7 @@ import java.util.UUID;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
-public class FinAppTests {
+public class FinAppIT {
 
   private static SpannerDaoInterface JDBCDao;
   private static SpannerDaoInterface JavaDao;
--- a/src/test/java/com/google/finapp/FinAppTests.java
+++ b/src/test/java/com/google/finapp/FinAppTests.java
@@ -39,6 +39,7 @@ public class FinAppTests {
 
   @BeforeClass
   public static void setUpTests() {
+    // set all IDs according to emulator setup
     String spannerProjectId = "test-project";
     String spannerInstanceId = "test-instance";
     String spannerDatabaseId = "test-database";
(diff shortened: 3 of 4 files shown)


src/test/java/com/google/finapp/FinAppTests.java
```diff
@@ -0,0 +1,93 @@
...
+    JDBCDao = new SpannerDaoJDBCImpl(spannerProjectId, spannerInstanceId, spannerDatabaseId);
+    SpannerOptions spannerOptions = SpannerOptions.getDefaultInstance();
+    Spanner spanner = spannerOptions.toBuilder().build().getService();
+    databaseClient = spanner
+        .getDatabaseClient(DatabaseId.of(spannerProjectId, spannerInstanceId, spannerDatabaseId));
+    JavaDao = new SpannerDaoImpl(databaseClient);
+  }
+
+  @Test
+  public void createAccountTest() throws SpannerDaoException {
```
ngcgarcia, line 55: I structured this test to handle both JDBC and Java implementations. This allows us to reduce a bit of repeated code (when the database is read to check results), but might be bloating the function a bit. Thoughts on keeping the two implementations together for the test or separating them?

src/test/java/com/google/finapp/FinAppTests.java
```diff
@@ -0,0 +1,93 @@
...
+    ByteArray accountIdJDBC = UuidConverter.getBytesFromUuid(UUID.randomUUID());
+    ByteArray accountIdJava = UuidConverter.getBytesFromUuid(UUID.randomUUID());
```
ngcgarcia, line 57: I use UUIDs for the sake of convenience, but this can be changed. 

UUIDs make it so that the database doesn't need to be cleared between each tests. If we had some set IDs, then running the tests twice in a row could cause the second tests to pass if the first did--even if something broke the code--since the test would query the info placed in the database prior.

One other alternative is to clear the database in the setup (and after the tests). This would make it so that we can have set database IDs. However, clearing the database might be undesirable for the user--although the test data we put in would technically "corrupt" the data in the database anyways.

src/test/java/com/google/finapp/FinAppTests.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.UUID;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class FinAppTests {
+
+  private static SpannerDaoInterface JDBCDao;
+  private static SpannerDaoInterface JavaDao;
+  private static DatabaseClient databaseClient;
```
ngcgarcia, line 38: This database client is passed into the Java client `SpannerDaoImpl` to use for its own mutations and used by the test suite to check the database for correct modifications. I think this should be fine, but please let me know if not.

biancamacias: im not sure if this helps, but maybe [this](https://kkjavatutorials.com/how-to-mock-interface-using-mockito-example/) is worth checking out

README.md
```diff
@@ -36,3 +36,10 @@ for client libraries to work.
     $ grpc_cli call localhost:8080 CreateCustomer \
         "name: 'google' address: 'amphitheatre pkwy'"
     ```
+
```
ngcgarcia, line 39: I can also delete these instructions; they might not be necessary. Also, we might change how we run tests soon.

README.md
```diff
@@ -36,3 +36,10 @@ for client libraries to work.
...
+## How to run the application tests
+
+1. Set up the emulator as described in #1 above.
+2. Replace the variables in `FinAppTests.setUpTests` according to your chosen instance, project, and database IDs.
+> The default variables should work for instance and database. The project variable must be changed based on `your-project-id` from `gcloud config set project your-project-id`. If this exact command was used, use `"your-project-id"`.
```
voulg, line 44: Optional:

How about we update the instructions above to use: ["test-project", "test-instance", "test-database"] and then remove this line? Actually it's weird that we use test-X but your-project-id...

In a separate PR we could move these to env variables and re-write the instructions to set the env

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private static SpannerDaoInterface JDBCDao;
+  private static SpannerDaoInterface JavaDao;
+  private static DatabaseClient databaseClient;
+
+  static {
+    // set all IDs according to emulator setup
+    String spannerProjectId = "test-project";
+    String spannerInstanceId = "test-instance";
+    String spannerDatabaseId = "test-database";
```
voulg, line 45: No action required:
As we discussed in our 1:1 let's consider replacing this with IntegrationTestEnv (https://github.com/googleapis/java-spanner/blob/47f59655acaf0fcf5848039e6b4881b80cdceb5c/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java#L40) that would come with a few more advantages.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+      return JDBCDao;
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"java client", "jdbc"})
+  public void createAccountTest(String impl) throws SpannerDaoException {
+    SpannerDaoInterface spannerDao = getImpl(impl);
+    ByteArray accountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
```
voulg, line 66: No action required:
Can you file an issue to change the interface of the CreateX commands to return the UUID instead of taking it as input?

voulg (approved):
LGTM, there are multiple improvements we could do but we can do them in separate PRs.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -0,0 +1,86 @@
...
+    spannerDao.createAccount(
+        accountId,
```
biancamacias, line 68:
```suggestion
        UuidConverter.getBytesFromUuid(UUID.randomUUID()),
```

src/test/java/com/google/finapp/FinAppIT.java
biancamacias, line 66:
```suggestion
```

biancamacias (approved):
this is a really solid test! it's very readable (i know exactly what its doing, can tell how its getting set up). i think this is a good base and will be so helpful for me when i start writing my own tests.

README.md
ngcgarcia: Documentation updated. See #26 about env variables

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/25

ngcgarcia: `accountId` is reused later to read from the database, so we must define it as a variable here.

Issue #25, opened 2021-08-03
Update method names and return UUIDs


biancamacias:
## Issue
currently, the RPC request/emthod names should be renamed (AddAccountForCustomer maybe not the best name, maybe say `request` in the name somewhere for the requests). also, the interface methods should return UUIDs.
## Next Steps
- [ ] update rpc request/method names (and add responses now?)
- [ ] `AddAccountForCustomer` -> `CreateCustomerRoleForAccount`? (in `FinAppService`)
- [ ] `createCustomer` should return `customerId`
- [ ] `createAccount` should return `accountId`
- [ ] `addAccountForCustomer` shoulder return `roleId`
this means changing the signatures across tests, interfaces, implementations, etc.

PR #27, opened 2021-08-03
Return UUIDs

biancamacias:
this PR edits some methods in `SpannerDaoInterface` and its implementations to now return its appropriate UUIDs and changes method names and message request names to be descriptive/to follow standard nomenclature (#25)

this PR accomplishes the following:

- [x] adds response messages to `service.proto`
- [x] edits `SpannerDaoInterface` to change some methods to a return type of `ByteArray`
- [x] edits both `SpannerDaoImpl` and `SpannerDaoJDBCImpl` to match the interface and return the UUIDs
- [x] edits `FinAppService` methods to now use rpc responses
- [x] renames `addAccountForCustoemr` to `createCustomerRole`
- [x] renames rpc requests to use `Request` in the name


src/main/proto/service.proto
```diff
@@ -24,18 +24,22 @@ option java_outer_classname = "ServiceProto";
 
 service FinApp {
 
-  rpc CreateCustomer(Customer) returns (.google.protobuf.Empty) {}
+  rpc CreateCustomer(Customer) returns (CustomerResponse) {}
```
voulg, line 27: I think it would be better to rename to:
CreateCustomerResponse
CreateAccountResponse
CreateCustomerRoleResponse

I'd expect a message that is called CustomerResponse to have a customer response...

src/main/proto/service.proto
```diff
@@ -24,18 +24,22 @@ option java_outer_classname = "ServiceProto";
...
 
-  rpc CreateAccount(Account) returns (.google.protobuf.Empty) {}
+  rpc CreateAccount(Account) returns (AccountResponse) {}
 
-  rpc AddAccountForCustomer(CustomerRole) returns (.google.protobuf.Empty) {}
+  rpc AddAccountForCustomer(CustomerRole) returns (CustomerRoleResponse) {}
```
voulg, line 31: How about renaming the AddCountForCustomer? I think we were discussing about:
CreateCustomerRoleRequest or something along these lines?
Will this be done in a separate PR?

src/main/proto/service.proto
```diff
@@ -24,18 +24,22 @@ option java_outer_classname = "ServiceProto";
...
 }
 
 message Customer {
   optional string name = 1;
   optional string address = 2;
 }
 
+message CustomerResponse {
+  optional bytes id = 1;
```
voulg, line 40: Nit: I'd rather call these customer_id, account_id, role_id to be even clearer.

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -25,7 +25,8 @@
   /**
    * Inserts a new row to the Customer table in the database.
    */
-  void createCustomer(ByteArray customerId, String name, String address) throws SpannerDaoException;
+  ByteArray createCustomer(ByteArray customerId, String name, String address)
```
voulg, line 28: The interface is a bit weird: Why do we pass the customerId and also return it? The caller already knows it.

We should either:
1. Do not change the interface. The service creates one and passes it to DAO and the response.
2. Change the interface to return the Id, but then do not pass the Id as a parameter and the DAO generates it and returns it.

I think I'd prefer 1 as the UUID generation does not change based on the DAO choice.

Thoughts?

ngcgarcia: ++ on option 1

src/main/proto/service.proto
biancamacias: yes im going to do that in this PR, i just wanted to get the OK on everything i've changed so far before i change the name.

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -33,24 +32,28 @@
...
+    CustomerResponse response = CustomerResponse.newBuilder()
+        .setId(ByteString.copyFrom(customerId.toByteArray())).build();
+    responseObserver.onNext(response);
     responseObserver.onCompleted();
   }
 
   @Override
-  public void createAccount(Account account, StreamObserver<Empty> responseObserver) {
+  public void createAccount(Account account, StreamObserver<AccountResponse> responseObserver) {
+    ByteArray accountId;
```
ngcgarcia, line 54: Here, you could assign `accountId` to the UUID and then remove the assignment below. Then, make the SpannerDao `createAccount`s return nothing.

src/main/java/com/google/finapp/SpannerDaoInterface.java
biancamacias: yes, that makes sense! i think i misunderstood where to return the UUID, haha, ill change this back.

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -68,15 +71,18 @@ public void createAccount(Account account, StreamObserver<Empty> responseObserve
...
+        .setId(ByteString.copyFrom(accountId.toByteArray())).build();
+    responseObserver.onNext(response);
     responseObserver.onCompleted();
   }
 
   @Override
   public void addAccountForCustomer(CustomerRole role,
-      StreamObserver<Empty> responseObserver) {
+      StreamObserver<CustomerRoleResponse> responseObserver) {
+    ByteArray roleId;
```
ngcgarcia, line 83: See other comment about assignment.

ngcgarcia (changes requested):

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -33,25 +32,31 @@
   }
 
   @Override
-  public void createCustomer(Customer customer, StreamObserver<Empty> responseObserver) {
+  public void createCustomer(Customer customer,
+      StreamObserver<CreateCustomerResponse> responseObserver) {
```
haseebajmal, line 36: nit: please make sure to format your code before submitting.

haseebajmal (approved):

src/main/proto/service.proto
```diff
@@ -16,26 +16,28 @@ syntax = "proto2";
...
-import "google/protobuf/empty.proto";
-
 option java_multiple_files = true;
 option java_package = "com.google.finapp";
 option java_outer_classname = "ServiceProto";
 
 service FinApp {
 
-  rpc CreateCustomer(Customer) returns (.google.protobuf.Empty) {}
+  rpc CreateCustomer(Customer) returns (CreateCustomerResponse) {}
```
ngcgarcia, line 25: Actually, can we standardize all request and response names? As in for the method `CreateCustomer` we would have `CreateCustomerRequest` (instead of just `Customer`) and `CreateCustomerResponse`. 

My reasoning is that future methods' requests can't be as easily simplified down to things like `Customer` or `Account`. For example, `MoveAccountBalance` requires 2 account ids and an amount (and is different from a "Transaction"). I will most likely label the request `MoveAccountBalanceRequest`, and I think it will be easier for users to understand and navigate this file if there are directly labeled request and responses for each RPC.

The downside to this is lack of reusability if there are multiple methods that share a request signature.

I'm not sure what the preferred nomenclature is for gRPC, but I think this would clarify things a bit.

Thoughts?

voulg: Using {Method}(Request|Response) is indeed standard practice and it makes it easier to evolve such protos in the long-run. I was thinking to propose it but decided against it for simplicity.

However, it seems you both like it, so go for it :) (with this or separate PR up to you)

src/main/proto/service.proto
voulg: Are we still changing the method name as part of this PR? I still see `AddAccountForCustomer`

biancamacias: yes, im gonna start changing all the of the names (method names and message names)

voulg (approved):
Thanks, remember to format before commit just in case.

ngcgarcia (approved):
Great PR! LGTM

Issue #26, opened 2021-08-03
Improve testing


ngcgarcia:
## Issue

As of now, testing directly modifies the database that the application uses. Furthermore, we explicitly define the project, instance, and database IDs in the tests, rather than possibly getting this information from the environment. 

We can also improve the tests to possibly use existing infrastructure to handle these problems and more. See: https://github.com/googleapis/java-spanner/blob/47f59655acaf0fcf5848039e6b4881b80cdceb5c/google-cloud-spanner/src/test/java/com/google/cloud/spanner/IntegrationTestEnv.java#L40

## Next Steps

- [x] Investigate IntegrationTestEnv to see if it or its ideas can be used for our tests
- [x] Look into better ways to set project, instance, and database IDs, possibly using the environment
- [x] Have tests possibly create and delete their own database

ngcgarcia:
Resolved by #28

Issue #11, opened 2021-07-19
Use logging in SpannerDaoImpl


ngcgarcia:
## Issue

In `SpannerDaoImpl`, we print out information related to database updates. This information should instead be put into logs via `java.util.logging`. 

## Steps to Reproduce the Problem

- [ ] Initialize the logger in the constructor
- [ ] Replace print statements with log inputs

PR #28, opened 2021-08-04
Use IntegrationTestEnv

ngcgarcia:
Partially addresses #26 

This PR introduces `IntegrationTestEnv` into this app's tests. This testing environment is used by both java-spanner and java-spanner-jdbc (and possibly other repos) to set up an integration test environment. This includes initializing a test database and deleting it at the end of tests.  Tests can be ran with `mvn integration-test`.

We also get rid of the explicit assignment of variables for project, instance, and database ID, instead getting these variables from `IntegrationTestEnv`, which is influenced by both the environment in which the tests are run (including general `gcloud` setup) and the config in the Maven `pom.xml` file. 

A lot of the decisions made in this PR were influenced by [previous uses of `IntegrationTestEnv`](https://github.com/googleapis/java-spanner/search?q=integrationtestenv) and [this `pom.xml` file](https://github.com/googleapis/java-spanner-jdbc/blob/6d545b617cd2ab21de88a77ce8e9ebe31cf088bf/pom.xml).

Another part of the PR is the removal of parametrized tests:

`ParameterizedTest` tests are run before the `BeforeClass` methods, as explained in [this StackOverflow forum](https://stackoverflow.com/questions/22391202/junit-parameterized-function-is-run-before-beforeclass-in-a-test-suite). Before, I  used a method to get around this by replacing my `BeforeClass` methods with a simple `static {}`, as described in the aforementioned forum. However, this had some strange interactions with the `ClassRule` (which was introduced with `IntegrationTestEnv`). 

Switching the `static` back to `BeforeClass` to ensure the order of `ClassRule` to `BeforeClass` to `Before` to tests, I had to rework the code to no longer use `ParameterizedTest`, instead opting to set the `SpannerDaoInterface` implementation through environment.

By default, the tests use the Java client implementation, but adding `-DSPANNER_USE_JDBC=true` to `mvn integration test` switches to the JDBC implementation.

Accomplished in this PR:

- update `pom.xml` to include necessary dependencies and set up `mvn integration-test` command
- change tests to use `IntegrationTestEnv`, handling necessary structural adjustments
- update README to have proper test command


src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,67 +19,120 @@
...
-    Spanner spanner = spannerOptions.toBuilder().build().getService();
-    databaseClient = spanner
-        .getDatabaseClient(DatabaseId.of(spannerProjectId, spannerInstanceId, spannerDatabaseId));
+  @ClassRule
+  public static IntegrationTestEnv env = new IntegrationTestEnv();
+
+  @BeforeClass
+  public static void setup() {
+    final RemoteSpannerHelper testHelper = env.getTestHelper();
+    db = testHelper.createTestDatabase(
```
ngcgarcia, line 54: It is not ideal to directly copy from the schema file. However, I had trouble finding either Cloud Spanner Java API or `IntegrationTestEnv`-related methods for using an `sdl` file to set up our databases programmatically. Most examples I saw used Cloud Spanner CLI (i.e. `gcloud spanner databases create test-database --ddl-file src/main/java/com/google/finapp/schema.sdl`).

Please let me know if you have any ideas for accomplishing this setup in Java or if this looks okay.

voulg: How about you load the schema from the file?

For example on how to do this, see here:
https://github.com/googleapis/java-spanner-jdbc/blob/6d545b617cd2ab21de88a77ce8e9ebe31cf088bf/src/main/java/com/google/cloud/spanner/jdbc/JdbcDatabaseMetaData.java#L51

btw, we should probably move the file in a resources folder, e.g: check out the files that are being loaded from the above:
https://github.com/googleapis/java-spanner-jdbc/tree/6d545b617cd2ab21de88a77ce8e9ebe31cf088bf/src/main/resources/com/google/cloud/spanner/jdbc

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,67 +19,120 @@
...
+            + "  StatementSummary BYTES(MAX) NOT NULL\n"
+            + ") PRIMARY KEY (AccountId, StatementId),\n"
+            + "  INTERLEAVE IN PARENT Account ON DELETE CASCADE");
+    final String databaseId = db.getId().getDatabase();
+    final String projectId = testHelper.getOptions().getProjectId();
+    final String instanceId = testHelper.getInstanceId().getInstance();
+    JDBCDao = new SpannerDaoJDBCImpl(projectId, instanceId, databaseId);
+    databaseClient = testHelper.getDatabaseClient(db);
     JavaDao = new SpannerDaoImpl(databaseClient);
+    System.out
```
voulg, line 101: Why not a logger instead of System.out?

ngcgarcia: #11

src/test/java/com/google/finapp/FinAppIT.java
biancamacias: +1 on putting all of this in the resources folder in its own file just for the purposes of testing.

biancamacias (commented):
solid! re-request my review when you've set up the resources folder, but this PR is looking great. i will try and figure out how to run this on my own to see if the `mvn integration-test` command works for me but looking good so far.

README.md
```diff
@@ -40,4 +40,4 @@ for client libraries to work.
 ## How to run the application tests
 
 1. Set up the emulator as described in #1 above.
```
haseebajmal, line 42: At some point we'd want to bring up the emulator as part of `mvn integration-test` right ?

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
-    SpannerOptions spannerOptions = SpannerOptions.getDefaultInstance();
-    Spanner spanner = spannerOptions.toBuilder().build().getService();
-    databaseClient = spanner
-        .getDatabaseClient(DatabaseId.of(spannerProjectId, spannerInstanceId, spannerDatabaseId));
+  @ClassRule
+  public static IntegrationTestEnv env = new IntegrationTestEnv();
+
+  @BeforeClass
+  public static void setup() {
+    final RemoteSpannerHelper testHelper = env.getTestHelper();
```
haseebajmal, line 59: It's generally discouraged to use final for all parameters and local variables. The dangers of sloppy reassignment of locals can be limited if you keep your methods short and focused.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
 
+@Category(IntegrationTest.class)
 public class FinAppIT {
 
   private static SpannerDaoInterface JDBCDao;
   private static SpannerDaoInterface JavaDao;
   private static DatabaseClient databaseClient;
+  private static Database db;
```
haseebajmal, line 52: static is good for True constants (When a final field is initialized to a hard-coded constant value) and ok-ish effective constants (technically mutable, the class is careful to never modify it once it is initialized). I would just drop static from all the variables here.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
-      return JavaDao;
-    } else {
-      return JDBCDao;
+  @AfterClass
+  public static void tearDown() {
+    db.drop();
+  }
+
+  @Test
+  public void createAccountTest() throws SpannerDaoException {
```
haseebajmal, line 80: A test's name should summarize the behavior it is testing. A good name describes both the actions that are being taken on a system and the expected outcome. It can get pretty verbose and that's ok.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
 
+@Category(IntegrationTest.class)
 public class FinAppIT {
```
haseebajmal, line 47: Is this really an integration test ? I imagined this to bring up a local grpc server and test via the public GRPC methods rather than building the spanner daos and testing those methods.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
+
+  @Test
+  public void createAccountTest() throws SpannerDaoException {
+    for (SpannerDaoInterface spannerDao : List.of(JavaDao, JDBCDao)) {
+      ByteArray accountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
+      spannerDao.createAccount(
+          accountId,
+          AccountType.UNSPECIFIED_ACCOUNT_TYPE /* = 0*/,
+          AccountStatus.UNSPECIFIED_ACCOUNT_STATUS /* = 0*/,
+          new BigDecimal(2));
```
haseebajmal, line 87: nit: Can we extract this into a variable - amount and reuse it here and in the assertion.

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: The `static` keyword seems to be required, since the later `@BeforeClass` method initializes the variables in a static context. (I get the error "non-static field cannot be referenced from a static context" if I try to remove the `static`)

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: This is a little unclear to me as well. It's not quite a unit test since it's running through both SpannerDao and the emulator. In my chats recently with @voulg and @c24t, we've been referring to this as an integration test, although it doesn't integrate GRPC but still tests the emulator and our code working together. So I believe it still merits this label because it's integrating the emulator and our code.

A full end-to-end test is ideal for a final product, but for this PR at least, we'll stick to these SpannerDao tests.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,104 @@
...
-    } else {
-      return JDBCDao;
+  @AfterClass
+  public static void tearDown() {
+    db.drop();
+  }
+
+  @Test
+  public void createAccountTest() throws SpannerDaoException {
+    for (SpannerDaoInterface spannerDao : List.of(JavaDao, JDBCDao)) {
```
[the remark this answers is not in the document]
ngcgarcia: Parameterization would be ideal, but, as my PR description mentions, there's some issues with using the parameterized tests. Do you have any suggestions for clean ways to get around the problems of parameterized tests happening before `BeforeClass` methods?

README.md
voulg: In the examples I've seen so far this is done before maven, e.g: https://github.com/googleapis/java-spanner-jdbc/blob/master/.github/workflows/integration-tests-against-emulator.yaml

It would probably be easier to follow this pattern

src/test/java/com/google/finapp/FinAppIT.java
voulg: Using com.google.cloud.spanner.IntegrationTestEnv should make this test able to run with either emulator or spanner. Moreover emulator (or real spanner) is considered part of the environment and is initialized outside the test (e.g: https://github.com/googleapis/java-spanner-jdbc/blob/master/.github/workflows/integration-tests-against-emulator.yaml), so IT is appropriate.

If the test was bringing emulator up internally I'd agree it's not a very clear cut.

src/test/java/com/google/finapp/FinAppIT.java
voulg: I'm ok with this solution, but for your consideration:
There's both `@BeforeClass` that is static and `@Before` that is non-static and runs after each test.
One way to think about it is:
- Objects that are expensive to construct and could be shared between tests can be done once on `@BeforeClass`. You get the static ugliness, but it's efficient. Creating a database is a good example of a potentially expensive operation
- Objects that are cheaper to construct can go to `@Before` and be non-static

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -19,68 +19,98 @@
...
   }
 
-  private SpannerDaoInterface getImpl(String impl) {
-    if (impl.equals("java client")) {
-      return JavaDao;
-    } else {
-      return JDBCDao;
+  @Test
+  public void createAccountTest() throws SpannerDaoException {
+    for (SpannerDaoInterface spannerDao : List.of(daoJava, daoJDBC)) {
```
voulg, line 75: Optional: Have you considered to have an env variable like spanner.testenv.use_jdbc and use it inside the test to chose which dao to instantiate?
Then in the `pom.xml` you could have two variants: integration-test-java, integration-test-jdbc that differ on how they set this environment variable.

src/test/java/com/google/finapp/FinAppIT.java
haseebajmal: I personally would recommend using @Before here. This would help get rid of static here and solve the parameterization issues mentioned elsewhere. The objects themselves shouldn't be very expensive to construct IMO - we can try comparing the run times of the two approaches perhaps. As long as the run time isn't grossly bad, IMO the readability benefits outweigh the performance aspects.

README.md
haseebajmal: I see. Yeah I guess as long as we don't have to manually run all the commands to bring up the emulator and set up the env variables etc., we should be good.

src/test/java/com/google/finapp/FinAppIT.java
haseebajmal: Acknowledged.

src/test/java/com/google/finapp/FinAppIT.java
voulg: My only concern would db DB creation when running against real spanner. DB Creation is actually non trivial for spanner backend and there are limits in the admin api rate of calls / etc...

Indeed in the JDBC tests, db creation is the only thing that happens on `@BeforeClass` https://github.com/googleapis/java-spanner-jdbc/blob/6d545b617cd2ab21de88a77ce8e9ebe31cf088bf/src/test/java/com/google/cloud/spanner/jdbc/ITAbstractJdbcTest.java#L93

README.md
ngcgarcia: Currently, we do run under the assumption that the emulator is being run in the background, which seems to be what the [java-docs-samples Spanner JDBC repository does](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/spanner/jdbc/src/test/java/com/example/spanner/jdbc/JdbcExamplesIT.java). It would be ideal to set up some sort of way to have the emulator start up with a command or something, but for the sake of time, I don't think this should be included in this PR.

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: Even `Before` is giving me the "non-static field cannot be referenced from a static context" error when I use it to try to initialize a non-static field.

ngcgarcia: Ignore what I said previously about the "non-static..." error. 

I think it's fair to initialize the SpannerDaoInterface in `@Before`, but I'm with @voulg  in that setting up the database should be done in `@BeforeClass`

ngcgarcia: I can remove the statics of fields initialized in `@Before`, but not `@BeforeClass`, since it seems that it is standard practice to put `static` in `@BeforeClass` method signatures. 

I'll put the database initialization in `@BeforeClass` and spanner Dao initialization in `@Before` (much less expensive, since you just pass in a client or strings)

ngcgarcia: Actually, there is an awkwardness in the fact that you need `testHelper` to grab the project and instance IDs, and `testHelper` isn't a field of the class. So to get around this, you would either need to make `testHelper` a static variable or the project and instance IDs. It would probably be less convoluted (with very slight performance gains) to just initialize the spanner dao in `BeforeClass`

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: I changed the tests to now utilize environment to choose the spanner dao impl. Setting up separate commands in `pom.xml` can be looked into in the future. 

running `mvn integration-test` now runs the Java client impl
and `mvn -DSPANNER_USE_JDBC=true integration-test` runs the JDBC impl

src/test/java/com/google/finapp/FinAppIT.java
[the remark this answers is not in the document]
ngcgarcia: See: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/28#discussion_r683747286

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -16,71 +16,104 @@
...
+  public static void setup() throws IOException {
+    RemoteSpannerHelper testHelper = env.getTestHelper();
+    db = testHelper
+        .createTestDatabase(
+            extractStatementsFromSDLFile("src/main/java/com/google/finapp/schema.sdl"));
+    databaseClient = testHelper.getDatabaseClient(db);
+    if (System.getProperty("SPANNER_USE_JDBC") == null || System.getProperty("SPANNER_USE_JDBC")
+        .equalsIgnoreCase("false")) {
+      spannerDao = new SpannerDaoImpl(databaseClient);
     } else {
```
ngcgarcia, line 66: This means that any input other than "false" will be considered as "true". I'm not sure if it's worth it to specifically make sure that the input is "true" and throw an error or something if not.

README.md
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/31

src/test/java/com/google/finapp/FinAppIT.java
haseebajmal: Acknowledged, SG. Thanks!

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -16,72 +16,112 @@
...
+
+  @AfterClass
+  public static void tearDown() {
+    db.drop();
+  }
+
+  @Test
+  public void createAccount_createsSingleValidAccount() throws Exception {
     ByteArray accountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
+    BigDecimal bigDecimalTwo = new BigDecimal(2);
```
haseebajmal, line 104: nit: s/bigDecimalTwo/amount

README.md
haseebajmal: Acknowledged. SG, thanks!

README.md
```diff
@@ -40,4 +40,4 @@ for client libraries to work.
...
-2. Run `mvn test`.
+2. Run `mvn integration-test`.
```
haseebajmal, line 43: I see some logic now in the test which uses a system property for JDBC. Does executing mvn integration-test run both jdbc and non-jdbc versions of the test ? Please update the README as appropriate.

haseebajmal (approved):
Please resolve all outstanding comments before merging.

src/test/java/com/google/finapp/FinAppIT.java
voulg: sgtm

voulg (approved):

README.md
ngcgarcia: They need to be run separately at the moment. The README now contains instructions for how to do so.

biancamacias (approved):

Issue #19, opened 2021-07-30
Finish implementing RPCs in SpannerDaoInterface


biancamacias:
## Issue
2 methods are included in both implementations but are not yet implemented in `FinAppService` or `service.proto`

these methods include:
- `addAccountForCustomer`
- `moveAccountBalance`

## Next Steps
in order to finish implementing these, we need to:
- [ ] add the rpc method to `service.proto` with its corresponding message request
- [ ] include the method in `FinAppService` by calling the method in `SpannerDaoInterface` with its appropriate parameters

ngcgarcia:
#20 addressed `addAccountForCustomer` (now `CreateCustomerRole`)

PR #30, opened 2021-08-06
Create RPC for MoveAccountBalance

ngcgarcia:
This PR partially addresses #19 by creating the RPC for `MoveAccountBalance`.

`MoveAccountBalance` takes a `fromAccountId`, `toAccountId`, and `amount` and transfers the amount between the two accounts. It returns the balance of the two accounts.

Accomplished in this PR:
- add RPC, request, and response to proto
- change `SpannerDaoInterface` and implementations to now return a mapping of account ID to balance
- add method to `FinAppService`

Example outputs:
`fromAccount` (`\372m\334\000\220^E\235\235\315\247\353\376\252\277\035` but simplified to `<fromAccount>` below): $5000
`toAccount` (`0\240\350\223\257\264B\205\241\004Gc\263\275\355J` but simplified to `<toAccount>` below): $5000

```
$ grpc_cli call localhost:8080 MoveAccountBalance "from_account_id: '<fromAccount>' to_account_id: '<toAccount>' amount: '500'" --channel_creds_type=insecure
connecting to localhost:8080
from_account_id_balance: "4500"
to_account_id_balance: "5500"
Rpc succeeded with OK status

$ grpc_cli call localhost:8080 MoveAccountBalance "from_account_id: '<fromAccount>' to_account_id: '<toAccount>' amount: '50000'" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Cannot transfer amount greater than original balance. fromAccount balance: 4500, amount: 50000

$ grpc_cli call localhost:8080 MoveAccountBalance "from_account_id: '<fromAccount>' to_account_id: '<toAccount>' amount: '4500'" --channel_creds_type=insecure
connecting to localhost:8080
from_account_id_balance: "0"
to_account_id_balance: "10000"
Rpc succeeded with OK status

$ grpc_cli call localhost:8080 MoveAccountBalance "from_account_id: '<fromAccount>' to_account_id: '<toAccount>' amount: 'x03'" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Invalid amount - x03. Expected a NUMERIC value

$ grpc_cli call localhost:8080 MoveAccountBalance "from_account_id: '<fromAccount>' to_account_id: '<toAccount>' amount: '-23'" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Amount transferred cannot be negative. amount: -23
```


src/main/proto/service.proto
```diff
@@ -29,6 +29,8 @@ service FinApp {
   rpc CreateAccount(Account) returns (.google.protobuf.Empty) {}
 
   rpc AddAccountForCustomer(CustomerRole) returns (.google.protobuf.Empty) {}
+
+  rpc MoveAccountBalance(MoveAccountBalanceRequest) returns (.google.protobuf.Empty) {}
```
ngcgarcia, line 33: No information needs to be returned other than if the transaction succeeded or not. This is currently communicated through RPC status ("RPC failed..." and "RPC succeeded..." messages), but the success/failure status could be made into a gRPC message if this is preferred.

ngcgarcia: Another possible output is updated account balances for accounts (something like fromAccountIdBalance, toAcccountIdBalance). Depends on how much logic we plan on offloading on client versus server.

biancamacias: +1 on returning the updated account balances

biancamacias (commented):
i do think the method should return something to prove it was successful, such as the updated account balances. it would probably look something like
`message MoveAccountBalanceResponse {
   optional string amount_transferred = 1;
   optional string to_account_balance = 2;
   optional string from_account_balance = 3;
}` or something similar

ngcgarcia:
> i do think the method should return something to prove it was successful, such as the updated account balances. it would probably look something like
> `message MoveAccountBalanceResponse { optional string amount_transferred = 1; optional string to_account_balance = 2; optional string from_account_balance = 3; }` or something similar

My plan is to just return the two balances. The user should know the amount that they transferred since they just sent that. And then an error message will be returned if the transfer didn't go through.

pom.xml
```diff
@@ -121,8 +121,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>8</source>
-                    <target>8</target>
+                    <source>9</source>
```
voulg, line 124: Why did this change? Is it relevant to this PR?

src/main/proto/service.proto
```diff
@@ -69,3 +71,14 @@ message CreateCustomerRoleRequest {
 message CreateCustomerRoleResponse {
   optional bytes role_id = 1;
 }
+
+message MoveAccountBalanceRequest {
+  optional bytes from_account_id = 1;
+  optional bytes to_account_id = 2;
```
haseebajmal, line 77: This applies to existing methods as well, so it's likely a tangential discussion: Bytes are a bit unwieldy when you are trying to locally test stuff out and adds some complexity in client side code. Should we consider changing these to int64. The storage format can still be Bytes.

@voulg What do you think ?

src/main/proto/service.proto
```diff
@@ -69,3 +71,14 @@ message CreateCustomerRoleRequest {
...
+  optional string amount = 3;
```
haseebajmal, line 78: Can we add a comment describing the format of amount since it's a free-form string ?

src/main/proto/service.proto
```diff
@@ -69,3 +71,14 @@ message CreateCustomerRoleRequest {
...
+}
+
+message MoveAccountBalanceResponse {
+  optional string from_account_id_balance = 1;
+  optional string to_account_id_balance = 2;
+}
```
haseebajmal, line 84: nit: add blank line at end of file

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -112,7 +113,8 @@ public void createCustomerRole(
     }
   }
 
-  public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
+  public Map<ByteArray, BigDecimal> moveAccountBalance(
```
haseebajmal, line 116: Return ImmutableMap instead. In general: "For parameters, use the most general type that still captures the important and relevant semantic guarantees you expect (Example - List, Map). For return types, use the most specific type that you're comfortable with committing to and which explicitly reveals the semantics (Example - ImmutableList, ImmutableMap)."

src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -49,6 +50,6 @@ void createCustomerRole(
    * @param amount amount transferred from fromAccountId to toAccountId, must be less than or equal
    *     to fromAccountId's account balance, must be non-negative
    */
-  void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
-      throws SpannerDaoException;
+  Map<ByteArray, BigDecimal> moveAccountBalance(
```
haseebajmal, line 53: Can we document what we are returning in the Javadoc with @returns tag.

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -102,6 +103,30 @@ public void createCustomerRole(
...
+  public void moveAccountBalance(
+      MoveAccountBalanceRequest request,
+      StreamObserver<MoveAccountBalanceResponse> responseObserver) {
+    Map<ByteArray, BigDecimal> accountBalances;
+    ByteArray fromAccountId = ByteArray.copyFrom(request.getFromAccountId().toByteArray());
+    ByteArray toAccountId = ByteArray.copyFrom(request.getToAccountId().toByteArray());
+    try {
+      accountBalances =
+          spannerDao.moveAccountBalance(
+              fromAccountId, toAccountId, new BigDecimal(request.getAmount()));
```
voulg, line 116: new BigDecimal can throw if the 'amount' is not a valid numeric. The exception would be uncaught and the grpc client wouldn't get a nice error. This is pre-existing issue, we do that on createAccount as well. Consider creating an issue to better handle exceptions and ensure we send good messages to customers.

No need to fix with current PR as there are other similar issues.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
voulg: +1

@haseebajmal: I do find it a bit weird to pass a Map just for two values. I saw the long discussions on why Java doesn't provide a Pair. What would be your thoughts for using AbstractMap.SimpleImmutableEntry<BigDecimal,BigDecimal> for fromAmount, toAmount?

src/main/proto/service.proto
voulg: That would be nice, unfortunately UUID is 128 bit :(

I agree with this: https://stackoverflow.com/questions/36344826/how-do-i-represent-a-uuid-in-a-protobuf-message We should either use bytes, or strings. Strings are more readable and there's a standard to map UUIDs to strings, but longer.

We have a similar suggestion for how to encode UUIDs in Spanner and I went with BYTES for storage efficiency, but it's likely for the gRPC using string might be better. Thoughts?

haseebajmal: Ah I see. Yeah, I kind of like the human readable string for the RPC.

voulg: sgtm, Nick, Bianca, thoughts?

In any case changing to strings would require a new PR focused just in that, perhaps we can file an issue.

src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
haseebajmal: Hm I think both Pair and AbstractMap.SimpleImmutableEntry<BigDecimal,BigDecimal> are semantics-free and hence make code a bit hard to read - which one is from balance and which one is to. I agree Map is weird too.

Maybe we could simply return the MoveAccountBalanceResponse directly here ? Ideally I think we would have an intermediate proto to give a good separation, but for a simple application such as this, it probably doesn't matter much.

voulg: SG, let's go with ImmutableMap

src/main/proto/service.proto
biancamacias: right now i think we've been using bytes for any UUIDs. im definitely ok with changing it, but if we do, we should move this to a separate PR because this means changing a few methods in `FinAppService` that would crowd this PR.

ngcgarcia: Yeah, I'm all here for switching to strings in a separate PR. See https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/32

pom.xml
ngcgarcia: IntelliJ prompted me to switch this after making a change earlier. I reverted this change without any issues, so this seems unnecessary.

src/main/proto/service.proto
ngcgarcia: This will apply to "balance" of CreateAccountRequest, along with future RPC messages. Should we make some sort of comment at the top or something so as to not repeat this comment throughout the code, or would that be preferred?

src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -102,6 +103,30 @@ public void createCustomerRole(
...
+    } catch (SpannerDaoException e) {
+      responseObserver.onError(Status.fromThrowable(e).asException());
```
c24t, line 118: If the user gives us a negative or too-high transfer amount then `moveAccountBalance` throws an `IllegalArgumentException`.  Since that isn't caught here we get the same not-nice error as in @voulg's comment above:

```
Rpc failed with status code 2, error message:
```

Ideally we'd set the right gRPC error code ([3: `INVALID_ARGUMENT` ](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)) and a nice user-facing error message here.

Example failing calls that cause the same error:

```
grpc_cli call ... MoveAccountBalance 'from_account_id: "..." to_account_id: "..." amount: "50000"'
```
```
grpc_cli call ... MoveAccountBalance 'from_account_id: "..." to_account_id: "..." amount: "-500"'
```
```
grpc_cli call ... MoveAccountBalance 'from_account_id: "..." to_account_id: "..." amount: "nop"'
```

src/main/proto/service.proto
c24t: One option is to use the same format as for spanner numeric types: http://cloud/spanner/docs/reference/rpc/google.spanner.v1 (search for "NUMERIC"):

> Encoded as string, in decimal format or scientific notation format.Decimal format:[+-]Digits[.[Digits]] or[+-][Digits].DigitsScientific notation:[+-]Digits[.[Digits]][ExponentIndicator[+-]Digits] or[+-][Digits].Digits[ExponentIndicator[+-]Digits](ExponentIndicator is "e" or "E")

I haven't checked if this is compatible with `BigDecimal` though.

Anyway probably worth putting a separate short comment on each actually-a-decimal field and then possibly a longer comment at the top about decimals in general.

src/main/java/com/google/finapp/FinAppService.java
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/33

src/main/proto/service.proto
```diff
@@ -69,3 +71,14 @@ message CreateCustomerRoleRequest {
...
+
+message MoveAccountBalanceRequest {
+  optional bytes from_account_id = 1;
+  optional bytes to_account_id = 2;
+  optional string amount = 3;
+}
+
+message MoveAccountBalanceResponse {
+  optional string from_account_id_balance = 1;
+  optional string to_account_id_balance = 2;
```
c24t, line 83: IMO an `{account_id: balance}` map would be more useful here, but not a blocking comment.

c24t (commented):
Logic LGTM but I think we could add some error handling and a test.

src/main/proto/service.proto
ngcgarcia: Since there's only 2, then maybe it's best to keep it simple? As far as how clients interact with this, if they use something like the Java gRPC client, it might just be simpler to do something like `response.getFromAcccountIdBalance` then maybe something like `response.get<map name>.get(fromAccountId)`. Not sure if this is even a correct interpretation/understanding of how maps in gRPC work and how the client would use them.

src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -125,26 +128,35 @@ public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, B
...
+                accountBalancesBuilder.put(toAccountId, newDestAmount);
                 return null;
               });
+      return accountBalancesBuilder.build();
     } catch (SpannerException e) {
+      // filter for IllegalArgumentExceptions thrown in lambda function above
+      Throwable cause = e.getCause();
+      if (cause instanceof IllegalArgumentException) {
+        throw new IllegalArgumentException(cause.getMessage());
+      }
```
ngcgarcia, line 159: Seems as though exceptions thrown in the lambda function end up being wrapped by `SpannerException`. The point of the filter is to give more specific/useful feedback for the user. However, this filtering can be moved to `FinAppService`, which is what calls this function. I chose to filter here, though, so you can see directly where the `IllegalArgumentException` is coming from and why we filter through `SpannerException`s

src/main/proto/service.proto
ngcgarcia: @haseebajmal I followed your example of using the string type (since `createAccount` balance is string). Is there a specific reason why you did string instead of numeric? I was assuming because there's an easy and direct way to convert between string and `BigDecimal`

haseebajmal: IMO using string (or any other alternatives Chris has suggested) is fine as long as you document well what values are acceptable - precision, +ve/-ve, currency etc., - so that a new user can easily specify the right values.

Repeating the documentation seems fine to me. Usually when browsing through RPC documentation, it's convenient to have this documentation easily accessible at field level. At least I don't usually look at file level comments unless I'm modifying the file.

haseebajmal: Regd. choice of string in createAccount: Yeah you're right. String to BigDecimal conversion is trivial so I chose that.

src/main/proto/service.proto
c24t: Sure, up to you. :)

src/main/java/com/google/finapp/SpannerDaoImpl.java
ngcgarcia: Actually, this needs to be in this location in order for the two spanner dao implementations to have identical expected outcomes for invalid inputs. Moving the filter elsewhere will break the tests as I have them setup--which expects `IllegalArgumentException` for illegal arguments. However, this isn't in the signature, so it's debatable if I should even be testing this.

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -124,4 +128,140 @@ public void createAccount_createsSingleValidAccount() throws Exception {
...
+          assertThat(resultSet.getBigDecimal(1)).isEqualTo(toAccountBalance.add(amount));
+          count++;
+        }
+      }
+      assertThat(count).isEqualTo(2);
+    }
+  }
+
+  @Test
+  public void moveAccountBalance_negativeAmount_throwsException() throws Exception {
```
ngcgarcia, line 191: We don't have throwing `IllegalArgumentException`s in the signature, so it's debatable of whether we should even test this or not. Or if we should add this to either the signature and/or the docstring

c24t: What about using a more specific custom checked exception instead? Then it's in the signature and docs, and gives us a more specific error.

It's possible that `SpannerDaoImpl.moveAccountBalance` could throw an `IllegalArgumentException` that isn't because of a bad argument from the end user, but because of some bad logic handling the request. But if the request would make the balance go negative that's clearly a bad argument from the user and we should send back the right gRPC status code (3).

ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/34

src/main/java/com/google/finapp/SpannerDaoImpl.java
ngcgarcia: #34

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -124,4 +128,140 @@ public void createAccount_createsSingleValidAccount() throws Exception {
...
+                "Account",
+                KeySet.newBuilder()
+                    .addKey(Key.of(fromAccountId))
+                    .addKey(Key.of(toAccountId))
+                    .build(),
+                Arrays.asList("AccountId", "Balance"))) {
+      int count = 0;
+      while (resultSet.next()) {
+        if (resultSet.getBytes(0).equals(fromAccountId)) {
+          assertThat(resultSet.getBigDecimal(1)).isEqualTo(fromAccountBalance.subtract(amount));
```
c24t, line 179: [`getBigDecimalList`](https://googleapis.dev/java/google-cloud-spanner/6.6.1/com/google/cloud/spanner/StructReader.html#getBigDecimalList-int-) might make your life easier here.

c24t (approved):
LGTM if you want to solve #34 in another PR.

c24t (approved):
LGTM again after 7332abb

src/test/java/com/google/finapp/FinAppIT.java
ngcgarcia: If I'm understanding how to use this function correctly, this gets the list of BigDecimals from a column, but we still need to retain the row relationship between accountId and balance and look for the specific entry.

PR #35, opened 2021-08-11
Add CreateTransactionForAccount method

ngcgarcia:
This PR adds the `CreateTransactionForAccount` method to all parts of the application, including `SpannerDaoInterface` and its implementations, the protobuf, and `FinAppService`. It also adds tests for the Spanner Dao methods. 

`<account1>` balance: 5000

```
$ grpc_cli call localhost:8080 CreateTransactionForAccount "account_id: '<account1>' amount: '500' is_credit: true" --channel_creds_type=insecure
connecting to localhost:8080
new_balance: "4500"
Rpc succeeded with OK status

$ grpc_cli call localhost:8080 CreateTransactionForAccount "account_id: '<account1>' amount: '1' is_credit: false" --channel_creds_type=insecure
connecting to localhost:8080
new_balance: "4501"
Rpc succeeded with OK status

$ grpc_cli call localhost:8080 CreateTransactionForAccount "account_id: '<account1>' amount: '5000' is_credit: true" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Account balance cannot be negative. original account balance: 4501, amount to be removed: 5000

$ grpc_cli call localhost:8080 CreateTransactionForAccount "account_id: '<account1>' amount: 'x' is_credit: true" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Invalid amount - x. Expected a NUMERIC value

$ grpc_cli call localhost:8080 CreateTransactionForAccount "account_id: '<account1>' amount: '-34' is_credit: true" --channel_creds_type=insecure
connecting to localhost:8080
Rpc failed with status code 3, error message: Amount transferred cannot be negative. amount: -34
```


src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -149,6 +151,64 @@ public void moveAccountBalance(ByteArray fromAccountId, ByteArray toAccountId, B
...
 
+  @Override
+  public BigDecimal createTransactionForAccount(
+      ByteArray accountId, BigDecimal amount, boolean isCredit) throws SpannerDaoException {
+    if (amount.signum() == -1) {
+      throw new IllegalArgumentException(
+          String.format("Amount transferred cannot be negative. amount: %s", amount.toString()));
+    }
+    try {
+      List<BigDecimal> newBalanceList = new ArrayList();
```
ngcgarcia, line 162: I could use some suggestions for this. The reason I use a list here is because of an issue that came up from the lambda function below. Initializing `newBalance` here instead of inside the lambda caused an error: "variable used in lambda expression should be final or effectively final". However, I can't access the `newBalance` variable outside of the lambda unless I define it outside the lambda. 

I got around this in the `moveAccountBalance` PR by [using an `ImmutableMap.Builder`](https://github.com/ngcgarcia/cloud-spanner-samples/blob/moveaccountbalance-rpc/src/main/java/com/google/finapp/SpannerDaoImpl.java#L119) that I modified within the lambda since we needed a map anyways and you can call methods on objects initialized outside the lambda within the lambda. This was the inspiration behind the list used here.

Let me know if:
- you have a suggestion for a better object to use instead of a list
- you find another way around this lambda function problem

src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -124,4 +129,154 @@ public void createAccount_createsSingleValidAccount() throws Exception {
...
+                .to(Value.COMMIT_TIMESTAMP)
+                .build()));
+    BigDecimal outputNewBalance =
+        spannerDao.createTransactionForAccount(accountId, amount, isCredit);
+    assertThat(outputNewBalance).isEqualTo(oldAccountBalance.subtract(amount));
+    try (ReadOnlyTransaction transaction = databaseClient.readOnlyTransaction();
+        ResultSet transactionResultSet =
+            transaction.read(
+                "TransactionHistory",
+                KeySet.all(),
```
ngcgarcia, line 160: We get `all` here instead of a specific key because the "TransactionHistory" table is keyed under timestamp and account id (have to include both for a `key`), and we don't know the exact time of the commit. I couldn't figure out a method with these keys/keysets for setting the account id but allowing the timestamp to be anything, so I just grab everything.

biancamacias (approved):

src/main/java/com/google/finapp/SpannerDaoImpl.java
c24t: Can you just return the balance from the lambda?

```diff
diff --git a/src/main/java/com/google/finapp/SpannerDaoImpl.java b/src/main/java/com/google/finapp/SpannerDaoImpl.java
--- a/src/main/java/com/google/finapp/SpannerDaoImpl.java
+++ b/src/main/java/com/google/finapp/SpannerDaoImpl.java
@@ -27,8 +27,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.List;

 final class SpannerDaoImpl implements SpannerDaoInterface {

@@ -159,43 +157,44 @@ final class SpannerDaoImpl implements SpannerDaoInterface {
           String.format("Amount transferred cannot be negative. amount: %s", amount.toString()));
     }
     try {
-      BigDecimal finalBalance = databaseClient
-          .readWriteTransaction()
-          .run(
-              transaction -> {
-                // Get account balances.
-                ResultSet resultSet =
-                    transaction.read(
-                        "Account",
-                        KeySet.singleKey(Key.of(accountId)),
-                        ImmutableList.of("Balance"));
-                BigDecimal oldBalance = null;
-                while (resultSet.next()) {
-                  oldBalance = resultSet.getBigDecimal("Balance");
-                }
-                if (oldBalance == null) {
-                  throw new IllegalArgumentException(
-                      String.format("Account not found: %s", accountId.toString()));
-                }
-                BigDecimal newBalance;
-                if (isCredit) {
-                  newBalance = oldBalance.subtract(amount);
-                } else {
-                  newBalance = oldBalance.add(amount);
-                }
-
-                if (newBalance.signum() == -1) {
-                  throw new IllegalArgumentException(
-                      String.format(
-                          "Account balance cannot be negative. original account balance: %s, amount to be removed: %s",
-                          oldBalance.toString(), amount.toString()));
-                }
-                transaction.buffer(
-                    ImmutableList.of(
-                        buildUpdateAccountMutation(accountId, newBalance),
-                        buildInsertTransactionHistoryMutation(accountId, amount, isCredit)));
-                return newBalance;
-              });
+      BigDecimal finalBalance =
+          databaseClient
+              .readWriteTransaction()
+              .run(
+                  transaction -> {
+                    // Get account balances.
+                    ResultSet resultSet =
+                        transaction.read(
+                            "Account",
+                            KeySet.singleKey(Key.of(accountId)),
+                            ImmutableList.of("Balance"));
+                    BigDecimal oldBalance = null;
+                    while (resultSet.next()) {
+                      oldBalance = resultSet.getBigDecimal("Balance");
+                    }
+                    if (oldBalance == null) {
+                      throw new IllegalArgumentException(
+                          String.format("Account not found: %s", accountId.toString()));
+                    }
+                    BigDecimal newBalance;
+                    if (isCredit) {
+                      newBalance = oldBalance.subtract(amount);
+                    } else {
+                      newBalance = oldBalance.add(amount);
+                    }
+
+                    if (newBalance.signum() == -1) {
+                      throw new IllegalArgumentException(
+                          String.format(
+                              "Account balance cannot be negative. original account balance: %s, amount to be removed: %s",
+                              oldBalance.toString(), amount.toString()));
+                    }
+                    transaction.buffer(
+                        ImmutableList.of(
+                            buildUpdateAccountMutation(accountId, newBalance),
+                            buildInsertTransactionHistoryMutation(accountId, amount, isCredit)));
+                    return newBalance;
+                  });
       return finalBalance;
     } catch (SpannerException e) {
       // filter for IllegalArgumentExceptions thrown in lambda function above
```

PR #36, opened 2021-08-11
Change repo structure to support multiple modules

ngcgarcia:
This PR changes our repo to support multiple modules.

The new structure:
- cloud-spanner-samples
    - finapp
       - server
           - src 
           - pom
       - workload
           - src
           - pom

Now, the `finapp` artifact is the parent of two modules, `server` and `workload`. Building `finapp` also builds `server` and `workload`, and these two submodules can be built separately as well.

It is probably most useful to look at the actual branch then look at the files changed section.


ngcgarcia:
I moved all dependencies to live in the parent pom file. Thoughts?

pom.xml
```diff
@@ -1,207 +1,152 @@
...
+      <maven.compiler.target>1.7</maven.compiler.target>
+  </properties>
+        
+
+  <modules>
+    <module>server</module>
+    <module>workload</module>
+  </modules>
+
+  <dependencyManagement>
```
ngcgarcia, line 24: All dependencies live in the parent pom file (and not the children). Thoughts?

ngcgarcia:
Some questions:

- Where should the Dockerfile be? 
- Do any of the `yaml` files need to change?

pom.xml
haseebajmal: Not all dependencies/plugins should live in the parent pom file, but only the shared ones.We would still need to refer to these dependencies/plugins from the child file, but version management happens in a common way in the parent.

I found https://devflection.com/posts/2020-04-12-maven-part-3/, which is a nice blog that explains these concepts of inheritance and aggregation.

haseebajmal:
> Some questions:
> 
> * Where should the Dockerfile be? Should its contents be editted to have `server` instead of `finapp`?
> * Do any of the `yaml` files need to change?

- Dockerfile and the gke yaml files should live in the server module.
- It should be fine to change the name of the jar in DockerFile from finapp to finapp-server. But I would just leave it untouched for now.

ngcgarcia:
I went ahead and made some changes after reading  https://devflection.com/posts/2020-04-12-maven-part-3/. 

This includes:
- keeping what will probably be shared build plugins in the parent pom and then referencing these in the `server` pom (removing unnecessary configuration, since this is done by the parent)
- removing other redundant/unnecessary parts of the `server` and `workload` pom files

I also decided to move all of the dependencies back to the `server` pom. If any of these dependencies are needed when making the workload, then we can move the dependency to the parent pom to share.

c24t (approved):
This looks right and the last round of changes LGTM, let's see what @haseebajmal thinks.

haseebajmal (approved):

PR #37, opened 2021-08-16
GetRecentTransactionsForAccount rpc

biancamacias:
This PR creates the last rpc outlined in the MVP. This executes a query of all the transactions based on timestamps and returns them in a repeated attribute. 

This PR does the following:

- [x] adds new method to `SpannerDaoInterface`
- [x] adds new method to `SpannerDaoImpl` and `SpannerDaoJDBCImpl` to return `ImmutableList` of `TransactionEntry` types
- [x] adds new rpc to `service.proto`
- [x] adds method to `FinAppService`

This PR does not have:

- tests

### DIFF
--- a/src/main/java/com/google/finapp/FinAppService.java
+++ b/src/main/java/com/google/finapp/FinAppService.java
@@ -15,6 +15,8 @@
 package com.google.finapp;
 
 import com.google.cloud.ByteArray;
+import com.google.cloud.Timestamp;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import com.google.protobuf.ByteString;
@@ -135,6 +137,33 @@ final class FinAppService extends FinAppGrpc.FinAppImplBase {
         MoveAccountBalanceResponse.newBuilder()
             .setFromAccountIdBalance(accountBalances.get(fromAccountId).toString())
             .setToAccountIdBalance(accountBalances.get(toAccountId).toString())
+            .build();
+    responseObserver.onNext(response);
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void getRecentTransactionsForAccount(
+      GetRecentTransactionsForAccountRequest request,
+      StreamObserver<GetRecentTransactionsForAccountResponse> responseObserver) {
+    ImmutableList<TransactionEntry> transactionEntries;
+    ByteArray accountId = ByteArray.copyFrom(request.getAccountId().toByteArray());
+    Timestamp beginTimestamp =
+        Timestamp.ofTimeSecondsAndNanos(
+            request.getBeginTimestamp().getSeconds(), request.getBeginTimestamp().getNanos());
+    Timestamp endTimestamp =
+        Timestamp.ofTimeSecondsAndNanos(
+            request.getEndTimestamp().getSeconds(), request.getEndTimestamp().getNanos());
+    try {
+      transactionEntries =
+          spannerDao.getRecentTransactionsForAccount(accountId, beginTimestamp, endTimestamp);
+    } catch (SpannerDaoException e) {
+      responseObserver.onError(Status.fromThrowable(e).asException());
+      return;
+    }
+    GetRecentTransactionsForAccountResponse response =
+        GetRecentTransactionsForAccountResponse.newBuilder()
+            .addAllTransactionEntry(transactionEntries)
             .build();
     responseObserver.onNext(response);
     responseObserver.onCompleted();
--- a/src/main/java/com/google/finapp/SpannerDaoImpl.java
+++ b/src/main/java/com/google/finapp/SpannerDaoImpl.java
@@ -22,6 +22,7 @@ import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Mutation;
 import com.google.cloud.spanner.ResultSet;
 import com.google.cloud.spanner.SpannerException;
+import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TransactionContext;
 import com.google.cloud.spanner.Value;
 import com.google.common.collect.ImmutableList;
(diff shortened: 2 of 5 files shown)


server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -427,4 +429,49 @@ public void createTransactionForAccount_tooLargeAmount_throwsException() throws
...
+                .set("AccountType")
+                .to(AccountType.UNSPECIFIED_ACCOUNT_TYPE.getNumber())
+                .set("AccountStatus")
+                .to(AccountStatus.UNSPECIFIED_ACCOUNT_STATUS.getNumber())
+                .set("Balance")
+                .to(toAccountBalance)
+                .set("CreationTimestamp")
+                .to(Value.COMMIT_TIMESTAMP)
+                .build()));
+    ImmutableMap result = spannerDao.moveAccountBalance(fromAccountId, toAccountId, amount);
```
ngcgarcia, line 466: Using `SpannerDao.moveAccountBalance` will mean that this test will also be testing `moveAccountBalance`. Our original goal for this test suite was for these tests to be as close to unit as possible, only testing one `SpannerDao` method per test (hence why we manually insert accounts). However, for this function, it'll be quite the hassle to manually do the `moveAccountBalance`, and these are technically integration tests. So it might be okay to use  `SpannerDao.moveAccountBalance`, even though it's a bit contrary to our original goals.

Thoughts on manually doing the move versus using the `SpannerDao` function?

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -427,4 +429,49 @@ public void createTransactionForAccount_tooLargeAmount_throwsException() throws
...
+    ImmutableList<TransactionEntry> history =
+        spannerDao.getRecentTransactionsForAccount(
+            fromAccountId, Timestamp.MIN_VALUE, Timestamp.MAX_VALUE);
+    assertThat(history).hasSize(1);
+    assertThat(history.get(0)).isInstanceOf(TransactionEntry.class);
+    assertThat(history.get(0).getAccountId())
+        .isEqualTo(ByteString.copyFrom(fromAccountId.toByteArray()));
+    assertThat(history.get(0).getIsCredit()).isEqualTo(true);
+    assertThat(history.get(0).getAmount()).isEqualTo(amount.toString());
+  }
```
ngcgarcia, line 476: Can you add some a test where more than one transaction is involved? To make sure that the function is working for multiple transactions.

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -427,4 +429,49 @@ public void createTransactionForAccount_tooLargeAmount_throwsException() throws
         IllegalArgumentException.class,
         () -> spannerDao.createTransactionForAccount(accountId, amount, isCredit));
   }
+
+  @Test
+  public void getRecentTransactionsForAccount_valid() throws Exception {
```
ngcgarcia, line 434: I'd change the second part of the function name to be a bit more specific. Maybe something like `returnsSingleTransaction`, and a future one can be `returnsMultipleTransactions`.

ngcgarcia (commented):
Great PR!

server/src/main/java/com/google/finapp/SpannerDaoInterface.java
```diff
@@ -65,4 +67,15 @@ void createCustomerRole(
    */
   BigDecimal createTransactionForAccount(ByteArray accountId, BigDecimal amount, boolean isCredit)
       throws SpannerDaoException;
+
+  /**
+   * Uses accountId to fetch TransactionHistory based on range of timestamps.
+   *
+   * @param beginTimestamp timestamp for where query begins
+   * @param endTimestamp timestamp for where query ends
```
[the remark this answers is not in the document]
haseebajmal: +1. We should have good request validation in the FinappService layer.

In addition to Nick's suggestion, we should also think along the lines of what we do when one or both of the timestamps are not specified. We can assume reasonable defaults where possible (For eg., if endTimestamp is missing, use current time maybe).

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -427,4 +429,49 @@ public void createTransactionForAccount_tooLargeAmount_throwsException() throws
...
+    ByteArray fromAccountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
+    ByteArray toAccountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
+    BigDecimal fromAccountBalance = new BigDecimal(20);
+    BigDecimal toAccountBalance = new BigDecimal(0);
+    BigDecimal amount = new BigDecimal(10);
+    databaseClient.write(
+        ImmutableList.of(
+            Mutation.newInsertBuilder("Account")
```
haseebajmal, line 442: Can we extract this into a separate method and re-use. 

Also we should use non-UNSPECIFIED enum types.

server/src/test/java/com/google/finapp/FinAppIT.java
haseebajmal: Ah the testing philosophy discussions :)

IMO, this current structure is fine. The ultimate goal for testing is: If code is buggy, is there a test which fails and does the test directly point to the issue. So if moveAccountBalance is buggy, since we have separate tests for moveAccountBalance, we should be able to fairly quickly pinpoint that  moveAccountBalance is failing from the dedicated tests, even though this would fail as well.

On the goal itself : "Our original goal for this test suite was for these tests to be as close to unit as possible, only testing one SpannerDao method per test". In principle, our goal should be behaviors and not methods. Most often these two coincide, but if they don't it's fine.

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -427,4 +429,49 @@ public void createTransactionForAccount_tooLargeAmount_throwsException() throws
...
+    ImmutableMap result = spannerDao.moveAccountBalance(fromAccountId, toAccountId, amount);
+    ImmutableList<TransactionEntry> history =
+        spannerDao.getRecentTransactionsForAccount(
+            fromAccountId, Timestamp.MIN_VALUE, Timestamp.MAX_VALUE);
+    assertThat(history).hasSize(1);
+    assertThat(history.get(0)).isInstanceOf(TransactionEntry.class);
+    assertThat(history.get(0).getAccountId())
+        .isEqualTo(ByteString.copyFrom(fromAccountId.toByteArray()));
+    assertThat(history.get(0).getIsCredit()).isEqualTo(true);
+    assertThat(history.get(0).getAmount()).isEqualTo(amount.toString());
```
haseebajmal, line 475: I would just build the expected transaction proto here and do assertThat(history).containsExactly(expected_transaction1, expected_transaction2).

Also make sure we are testing the order of the transactions returned.

haseebajmal:
Also, while you're here, would you mind creating an issue to add pagination for this RPC ?

haseebajmal (approved):
LG. Let me know if you want me to take a second look.

server/src/main/java/com/google/finapp/SpannerDaoJDBCImpl.java
```diff
@@ -207,6 +211,42 @@ public BigDecimal createTransactionForAccount(
...
+      ResultSet resultSet = readStatement.executeQuery();
+      ImmutableList.Builder<TransactionEntry> transactionHistoriesBuilder = ImmutableList.builder();
+      while (resultSet.next()) {
+        transactionHistoriesBuilder.add(
+            TransactionEntry.newBuilder()
+                .setAccountId(ByteString.copyFrom(resultSet.getBytes("AccountId")))
+                .setEventTimestamp(
+                    com.google.protobuf.Timestamp.newBuilder()
+                        .setNanos(resultSet.getTimestamp("EventTimestamp").getNanos())
+                        .setSeconds(resultSet.getTimestamp("EventTimestamp").getSeconds()))
```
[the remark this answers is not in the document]
biancamacias: the result set types are different in the implementations, so the `.get` methods also return something different. when you do `resultSet.getTimestamp(...)` in the java implementation, you get a `com.google.cloud.Timestamp` which has a really nice `toProto()` method. in the jdbc implementation, you get a `java.sql.Timestamp`, so i had to use the Timestamp builder.

PR #38, opened 2021-08-17
Establish basic workload generator

ngcgarcia:
Blocked on #36 (and based on #36's branch)

This PR establishes the basic foundation for a workload generator. This current workload generator creates 200 accounts and then continuously sends `moveAccountBalance` requests for these accounts.

This can be ran in the `workload` directory with:
```
mvn -DskipTests package exec:java -Dexec.mainClass=com.google.finapp.WorkloadGenerator
```

Notes on the current generator:
1) requires the server to be ran separately
1) requires the database to be created beforehand (`$ gcloud spanner databases create test-database --ddl-file src/main/java/com/google/finapp/schema.sdl`)
1) only uses `createAccount` and `moveAccountBalance`
1) single threaded


c24t (commented):
Can we move these proto files up? They should even be shared across all impls, not just java.

ngcgarcia:
> Can we move these proto files up? They should even be shared across all impls, not just java.

Addressed in #38

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.protobuf.ByteString;
+import io.grpc.ManagedChannel;
+import io.grpc.StatusRuntimeException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class WorkloadClient {
+
+  private final ManagedChannel channel;
+  private static Logger logger = Logger.getLogger(WorkloadClient.class.getName());
```
ngcgarcia, line 13: I need to look into how to modify this for when the workload generator is multithreaded. For now, this works.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -0,0 +1,59 @@
...
+            .setType(type)
+            .setStatus(status)
+            .build();
+    try {
+      CreateAccountResponse response = blockingStub.createAccount(request);
+      logger.log(Level.INFO, String.format("Account created %s", response));
+      return response.getAccountId();
+    } catch (StatusRuntimeException e) {
+      logger.log(Level.SEVERE, String.format("Error creating account %s", request));
+      return null;
```
ngcgarcia, line 38: I have the code not error out when there's an exception creating an account. This is so that if there's some small error or something and only one or a small amount of `createAccount`s fails, the workload can continue regardless. 

This can be pretty easily changed to make the code fail automatically. Let me know your thoughts.

ngcgarcia: This `null` is intercepted in the `WorkloadGenerator` and essentially ignored.

workload/pom.xml
```diff
@@ -16,4 +16,77 @@
     <name>workload</name>
     <version>1.0-SNAPSHOT</version>
 
+    <dependencies>
```
ngcgarcia, line 19: #44

workload/src/main/proto/database.proto
```diff
@@ -0,0 +1,33 @@
+// Copyright 2021 Google LLC
```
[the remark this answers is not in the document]
ngcgarcia: Chris mentioned this too earlier, and I pushed this off into #43 for now since I couldn't find a quick solution to point the compiler to use proto files that aren't in the standard maven file structure (in `src/main/proto` of the module). However, I can look into fixing that in this PR if you'd prefer.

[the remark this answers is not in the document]
haseebajmal: I was looking into this the other day. So the way to do this would be:
1. Create a new maven module (serviceprotos) which hosts only the service protos. This would have its own pom with dependencies for grpc and protobuf. You can either put the proto file in src/main/proto in this module OR put it in the main folder updating the plugins protoSourceRoot attribute - https://www.xolstice.org/protobuf-maven-plugin/compile-mojo.html#protoSourceRoot
2. To finapp's (parent) pom, we add this serviceprotos module under <modules>
3. To both server and workload poms, we add the dependency:
```
        <dependency>
            <groupId>org.example</groupId>
            <artifactId>serviceprotos</artifactId>
            <version>1.0-SNAPSHOT</version>
        </dependency>
```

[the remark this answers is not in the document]
ngcgarcia: Ah, I was looking into this protoSourceRoot earlier. I think the plan for now is to just specify the filepath to the server module (and include it as a dependency) and remove the copies. 

I think a greater restructuring should happen in another PR. I'll copy your comment into #43.

Does this sound alright?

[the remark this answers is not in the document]
ngcgarcia: This change is reflected now in the most recent commit.

workload/src/main/java/com/google/finapp/WorkloadGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.finapp.CreateAccountRequest.Status;
+import com.google.protobuf.ByteString;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+public class WorkloadGenerator {
```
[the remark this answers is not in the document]
haseebajmal: Can we update the README file adding instructions on how to run the workload ?

workload/src/main/java/com/google/finapp/WorkloadGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+        ids.add(response);
+      }
+    }
+    Random random = new Random();
+    int numIds = ids.size();
+    if (numIds == 0) {
+      throw new RuntimeException("No accounts were created successfully.");
+    }
+    while (true) {
+      ByteString fromId = ids.get(random.nextInt(numIds));
```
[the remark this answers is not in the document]
ngcgarcia: This is just a first draft for the workload with pretty basic logic. As it is now, the workload will eventually (but will definitely take a _while_ to get there) start making some invalid transactions, which should also be supported eventually. The logic of the actual commands will be improved in future PRs.

workload/src/main/java/com/google/finapp/WorkloadGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+    int port = Integer.parseInt(args[1]);
+    ManagedChannel channel =
+        ManagedChannelBuilder.forAddress(addressName, port).usePlaintext().build();
+    List<ByteString> ids = new ArrayList<>();
+    for (int i = 0; i < 200; i++) {
+      ByteString response =
+          WorkloadClient.getWorkloadClient(channel)
+              .createAccount(
+                  "1000",
+                  CreateAccountRequest.Type.UNSPECIFIED_ACCOUNT_TYPE,
```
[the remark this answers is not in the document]
ngcgarcia: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/47

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -0,0 +1,154 @@
...
+      return parser.parse(options, args);
+    } catch (ParseException e) {
+      System.out.println(e.getMessage());
+      formatter.printHelp(
+          String.format(
+              "java -jar %s",
+              Paths.get("WorkloadMain.java").toAbsolutePath().normalize().toString()),
+          options);
+      System.exit(1);
+      return null;
```
ngcgarcia, line 151: The program terminates before this with `System.exit`  but the `return null` is necessary for the code to compile.

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  public static void main(String[] args) {
+    CommandLine cmd = parseArgs(args);
+    String addressName = cmd.getOptionValue("a");
+    int port;
+    int numAccounts;
+    try {
+      port = ((Number) cmd.getParsedOptionValue("p")).intValue();
+      numAccounts = ((Number) cmd.getParsedOptionValue("n")).intValue();
```
ngcgarcia, line 97: The signature of `getParsedOptionValue` returns an `Object`, but it is the type as defined for the `Option` (in this case, `Number`). 

See [these examples](https://www.tabnine.com/code/java/methods/org.apache.commons.cli.CommandLine/getParsedOptionValue) (especially the second-on) for common use cases.

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -0,0 +1,154 @@
...
+            .type(String.class)
+            .hasArg()
+            .build());
+
+    options.addOption(
+        Option.builder("p")
+            .longOpt("port")
+            .desc("server port")
+            .required(true)
+            .type(Number.class)
```
ngcgarcia, line 125: `Integer.class` is not allowed here, and `Number.class` is used instead. See: https://stackoverflow.com/questions/5585634/apache-commons-cli-option-type-and-default-value

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+
+/**
+ * A workload generator for the finance sample app that creates traffic by sending gRPC requests to
+ * the server.
+ */
+public final class WorkloadMain {
+  private static final String DEFAULT_ACCOUNT_BALANCE = "10000";
+  private static final String DEFAULT_TRANSFER_AMOUNT = "20";
```
biancamacias, line 41: not a need, but in addition to a default transfer amount, we could have some equation that takes a percentage or amount from the balance. like having the transfer amount be equal to `account_balance - (account_balance * 0.01)` to make the transfer amounts different every time or something similar.

biancamacias (approved):

PR #50, opened 2021-08-26
Make workload generator multithreaded

### LINKED ISSUE #54: Consider removing task queue in workload generator
See https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/50#discussion_r698908313 and https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/50#discussion_r699638483.

Consider removing the task queue in the workload generator and instead generating new tasks on demand, or using a `SynchronousQueue` instead of a `BlockingQueue` in the executor as described in https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html.

ngcgarcia:
This PR restructures the workload generator to be multithreaded. This change makes it so that clients run indefinitely, creating their own accounts and performing methods at random with those accounts. This removes the need for the `seedData` method of `WorkloadMain` since the account ids do not need to be generated beforehand.

Accomplished in this PR:
- change `WorkloadClient` to run indefinitely and generate random requests
- change `WorkloadMain` to create multiple clients according to `thread-count` argument
- added a commandline argument to adjust number of threads
- added javadocs

Fixes #54

### DIFF
--- a/workload/src/main/java/com/google/finapp/WorkloadClient.java
+++ b/workload/src/main/java/com/google/finapp/WorkloadClient.java
@@ -32,10 +32,11 @@ import java.util.logging.Logger;
  */
 public class WorkloadClient implements Runnable {
 
-  private final FinAppBlockingStub blockingStub;
   private static final Logger logger = Logger.getLogger(WorkloadClient.class.getName());
+  private static final int MAX_INITIAL_ACCOUNT_BALANCE = 20000;
   private final List<ByteString> ids;
   private final Random random = new Random();
+  private final FinAppBlockingStub blockingStub;
 
   private WorkloadClient(ManagedChannel channel) {
     this.blockingStub = FinAppGrpc.newBlockingStub(channel);
@@ -61,18 +62,16 @@ public class WorkloadClient implements Runnable {
         case 1:
           int idsSize = ids.size();
           int fromAcctIndex = random.nextInt(idsSize);
-          int toAcctIndex = random.nextInt(idsSize);
-          while (toAcctIndex == fromAcctIndex) {
+          int toAcctIndex;
+          do {
             toAcctIndex = random.nextInt(idsSize);
-          }
+          } while (toAcctIndex == fromAcctIndex);
           moveAccountBalance(
               ids.get(fromAcctIndex), ids.get(toAcctIndex), getRandomAmountFromRange(1, 200));
           break;
       }
     }
   }
-
-  private static final int MAX_INITIAL_ACCOUNT_BALANCE = 20000;
 
   private void addAccountWithRandomBalance() {
     ids.add(


workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -0,0 +1,139 @@
...
+  private final List<ByteString> ids;
+  private final Random random = new Random();
+
+  private WorkloadClient(ManagedChannel channel, ImmutableList<Task> tasks) {
+    this.blockingStub = FinAppGrpc.newBlockingStub(channel);
+    this.tasks = tasks;
+    this.ids = new ArrayList<>();
+  }
+
+  public static WorkloadClient getWorkloadClient(
```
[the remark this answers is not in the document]
ngcgarcia: Done.

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -33,75 +36,66 @@
...
  */
 public final class WorkloadMain {
-  private static final String DEFAULT_ACCOUNT_BALANCE = "10000";
-  private static final String DEFAULT_TRANSFER_AMOUNT = "20";
   private static final Logger logger = Logger.getLogger(WorkloadMain.class.getName());
 
   private static class WorkloadGenerator {
     private final ManagedChannel channel;
-    private final List<ByteString> ids = new ArrayList<>();
+    private static final int DEFAULT_MAX_QUEUE_SIZE = 100;
```
c24t, line 47: Since the number of concurrent workers is controlled by `threadCount` why keep so many tasks in the queue? Put another way, would we get the same performance with a queue size of 1?

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -0,0 +1,150 @@
...
+
+    void startSteadyLoad(int threadCount, int taskCount) {
+      ThreadPoolExecutor executor = (ThreadPoolExecutor) Executors.newFixedThreadPool(threadCount);
+      while (true) {
+        if (executor.getQueue().size() < DEFAULT_MAX_QUEUE_SIZE) {
+          ImmutableList<Task> tasks = generateRandomTasks(taskCount);
+          logger.log(Level.INFO, String.format("Tasks submitted %s", tasks.toString()));
+          executor.submit(WorkloadClient.getWorkloadClient(channel, tasks));
+        }
+      }
```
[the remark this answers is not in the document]
c24t: I agree that none of these policies solve the problem, and without some substantial restructuring it's better to just repeatedly check the queue size. Ideally we could get rid of the busy-wait and only create and `submit` a new task when space opens up on the queue.

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -33,75 +36,66 @@
...
-          logger.log(
-              Level.WARNING,
-              String.format("CreateAccount failed. Total count: %d", numFailedCreateAccounts));
+    void startSteadyLoad(int threadCount, int taskCount) {
+      ThreadPoolExecutor executor = (ThreadPoolExecutor) Executors.newFixedThreadPool(threadCount);
+      BlockingQueue<Runnable> queue = executor.getQueue();
+      while (true) {
+        if (queue.size() < DEFAULT_MAX_QUEUE_SIZE) {
+          ImmutableList<Task> tasks = generateRandomTasks(taskCount);
+          logger.log(Level.INFO, String.format("Tasks submitted %s", tasks.toString()));
```
c24t, line 63: Nit: shouldn't you log this after submitting the task? (Also confusing that this "task" is multiple tasks.)

workload/src/main/java/com/google/finapp/WorkloadMain.java
```diff
@@ -126,9 +120,18 @@ private static CommandLine parseArgs(String[] args) {
...
+            .desc("number of threads to use in thread pool")
+            .required(true)
+            .type(Number.class)
+            .hasArg()
+            .build());
+
+    options.addOption(
+        Option.builder("c")
+            .longOpt("task-count")
+            .desc("number of tasks (RPC methods) each thread performs")
```
c24t, line 134: Out of curiosity: since we keep starting new worker threads what's the difference between (1) using a limited number of tasks per worker and (2) using a limited number of workers that each generate infinitely many tasks?

c24t (commented):
Some suggestions for improvements in `WorkloadClient`, more speculative comments about executors in `WorkloadMain`.

Great to see this working, and the structure LGTM. I think it should be possible to write this without the busy-wait and the queued tasks but that can be a problem for another PR.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -14,28 +14,91 @@
...
+import java.util.Random;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-/** A gRPC client for the finance sample app. */
-public class WorkloadClient {
+/** A gRPC client for the finance sample app. Performs a series of tasks (gRPC methods) in order. */
+public class WorkloadClient implements Runnable {
 
   private final FinAppBlockingStub blockingStub;
```
c24t, line 32: Another interesting thing to consider: since we're reusing the channel across all threads and using blocking stubs the bottleneck here might be gRPC. I wonder if you get higher QPS with non-blocking stubs or separate channels per worker.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -0,0 +1,139 @@
...
+                getRandomAmountFromRange(0, 20000),
+                CreateAccountRequest.Type.CHECKING,
+                Status.ACTIVE));
+      }
+    }
+  }
+
+  public ByteString createAccount(
+      String balance, CreateAccountRequest.Type type, CreateAccountRequest.Status status)
+      throws StatusRuntimeException {
```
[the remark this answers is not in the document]
ngcgarcia: Acknowledged.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -14,28 +14,91 @@
...
+public class WorkloadClient implements Runnable {
 
   private final FinAppBlockingStub blockingStub;
   private static final Logger logger = Logger.getLogger(WorkloadClient.class.getName());
+  private final ImmutableList<Task> tasks;
+  private final List<ByteString> ids;
+  private final Random random = new Random();
 
-  private WorkloadClient(ManagedChannel channel) {
+  private WorkloadClient(ManagedChannel channel, ImmutableList<Task> tasks) {
```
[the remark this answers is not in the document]
ngcgarcia: I wanted to pass immutable objects to this client since it needs to be threadsafe. I now realize that the client is instantiated before passed to a thread, so this conversion from list to immutable list can happen during the client instantiation.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -14,28 +14,91 @@
...
+      switch (task) {
+        case CreateAccount:
+          ids.add(
+              createAccount(
+                  getRandomAmountFromRange(0, 20000),
+                  CreateAccountRequest.Type.CHECKING,
+                  Status.ACTIVE));
+          break;
+        case MoveAccountBalance:
+          ensureIdsPresent(2);
```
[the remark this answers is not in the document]
ngcgarcia: Done.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -14,28 +14,91 @@
...
+          List<ByteString> randomIds = getRandomUniqueIds(2);
+          moveAccountBalance(randomIds.get(0), randomIds.get(1), getRandomAmountFromRange(1, 200));
+          break;
+      }
+    }
+  }
+
+  private String getRandomAmountFromRange(int min, int max) {
+    return String.valueOf(random.nextInt(max - min) + min);
```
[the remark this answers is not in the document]
ngcgarcia: This function is used twice now (in `addAccountWithRandomBalance` and `run`). The function is only one line of code, but I think labeling this makes it much easier for the reader to understand.

workload/src/main/java/com/google/finapp/WorkloadMain.java
ngcgarcia: 1) Done
2) Are you talking about the naming of the enum in general? (name it `Tasks` instead of `Task`?)

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -14,28 +14,91 @@
...
+    }
+    List<ByteString> idsCopy = new ArrayList<>(ids);
+    List<ByteString> randomIds = new ArrayList<>();
+    for (int i = 0; i < numIds; i++) {
+      int index = random.nextInt(idsCopy.size());
+      randomIds.add(idsCopy.get(index));
+      idsCopy.remove(index);
+    }
+    return randomIds;
+  }
```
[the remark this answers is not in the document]
ngcgarcia: Done.

workload/src/main/java/com/google/finapp/WorkloadMain.java
ngcgarcia: The original intention was for each client to mock a customer with a finite number of tasks but doing (2) would be more efficient. I can rework the code to do (2) if this is preferred or make an issue for it to be done in a future PR. Thoughts?

workload/src/main/java/com/google/finapp/WorkloadMain.java
ngcgarcia: Since we have the small loop to check queue size, I would imagine that the queue would tend to stay near capacity. In this case, we could even consider a solution where we don't have a queue (but then we'd have to consider different logic for checking if we should add a new task). We could limit this queue size to something smaller regardless. I can set this to a smaller number and then create an issue for exploring different queue sizes and strategies, or I can look into this more now. Thoughts?

workload/src/main/java/com/google/finapp/WorkloadMain.java
c24t: Didn't mean for you to make any change for (2), just noting that "task" has a different meaning for the user and reviewer.

workload/src/main/java/com/google/finapp/WorkloadMain.java
c24t: I don't think you need to change it in this PR, just noting that (2) might be easier since you wouldn't need to queue anything, just run each worker in its own thread. FWIW since workers move balances between random accounts they don't really represent individual customers now.

workload/src/main/java/com/google/finapp/WorkloadMain.java
c24t: > I can set this to a smaller number and then create an issue for exploring different queue sizes and strategies, or I can look into this more now

I don't mean for this to be a blocking comment, this just looks to me like some evidence that we don't really need the task queue.

c24t: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/54

workload/src/main/java/com/google/finapp/WorkloadClient.java
ngcgarcia: #55

workload/src/main/java/com/google/finapp/WorkloadMain.java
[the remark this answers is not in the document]
ngcgarcia: #55

workload/src/main/java/com/google/finapp/WorkloadMain.java
ngcgarcia: #55

workload/src/main/java/com/google/finapp/WorkloadMain.java
ngcgarcia: Done.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -54,20 +102,19 @@ public ByteString createAccount(
...
             .setFromAccountId(fromAccountId)
             .setToAccountId(toAccountId)
             .build();
     try {
       MoveAccountBalanceResponse response = blockingStub.moveAccountBalance(request);
       logger.log(Level.INFO, String.format("Move made %s", response));
     } catch (StatusRuntimeException e) {
-      logger.log(Level.SEVERE, String.format("Error making move %s", request));
-      throw e;
+      logger.log(Level.INFO, String.format("Ignoring exception in moveAccountBalance: %s", e));
```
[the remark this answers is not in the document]
ngcgarcia: Will put in an issue or update https://github.com/GoogleCloudPlatform/cloud-spanner-samples/issues/53, unless you want me to address this now.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -17,30 +17,81 @@
...
   }
 
-  public ByteString createAccount(
-      String balance, CreateAccountRequest.Type type, CreateAccountRequest.Status status)
+  @Override
+  public void run() {
+    for (int i = 0; i < 2; i++) { // ensure that > 2 accounts exists for future methods
+      addAccountWithRandomBalance();
+    }
+    int numMethods = 2; // must be updated when new methods are added
```
ngcgarcia, line 54: For now, we use `numMethods` to describe the size of `{moveAccountBalance, createAccount}`, generate a random `int` to `switch` off of, and choose which method to run. This whole `run` method can be implemented in various ways, but I found this to be simple enough.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -17,30 +17,81 @@
...
+        case 0:
+          addAccountWithRandomBalance();
+          break;
+        case 1:
+          int idsSize = ids.size();
+          int fromAcctIndex = random.nextInt(idsSize);
+          int toAcctIndex = random.nextInt(idsSize);
+          while (toAcctIndex == fromAcctIndex) {
+            toAcctIndex = random.nextInt(idsSize);
+          }
```
c24t, line 67: Nit: do/while would save you a line here.

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -17,30 +17,81 @@
...
+          moveAccountBalance(
+              ids.get(fromAcctIndex), ids.get(toAcctIndex), getRandomAmountFromRange(1, 200));
+          break;
+      }
+    }
+  }
+
+  private static final int MAX_INITIAL_ACCOUNT_BALANCE = 20000;
```
c24t, line 75: Nit: move this up top?

workload/src/main/java/com/google/finapp/WorkloadClient.java
```diff
@@ -54,20 +105,26 @@ public ByteString createAccount(
...
       logger.log(Level.INFO, String.format("Move made %s", response));
     } catch (StatusRuntimeException e) {
-      logger.log(Level.SEVERE, String.format("Error making move %s", request));
-      throw e;
+      if (e.getStatus().equals(Status.INVALID_ARGUMENT)) {
+        logger.log(
+            Level.INFO,
+            String.format("Ignoring invalid argument error in moveAccountBalance: %s", e));
+      } else {
+        logger.log(Level.SEVERE, String.format("Unexpected error in moveAccountBalance: %s", e));
```
c24t, line 125: Good call on the error handling

c24t (approved):
LGTM! Much simpler than https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/50/commits/fe657a4517d043769b5404f9942c8f762e324569, nice work.

biancamacias (approved):

workload/src/main/java/com/google/finapp/WorkloadClient.java
ngcgarcia: Done.

PR #59, opened 2021-09-28
Make FinAppIT test be at the gRPC layer

voulg:
Motivation, among other things:
- Provides better coverage, already identified 2 bugs that I fixed as part of this PR
- Would allow moving common logic from doa implementation to FinAppService

### DIFF
--- a/server/src/main/java/com/google/finapp/FinAppService.java
+++ b/server/src/main/java/com/google/finapp/FinAppService.java
@@ -197,12 +197,16 @@ final class FinAppService extends FinAppGrpc.FinAppImplBase {
     } catch (SpannerDaoException e) {
       responseObserver.onError(Status.fromThrowable(e).asException());
       return;
+    } catch (IllegalArgumentException e) {
+      responseObserver.onError(Status.fromThrowable(e).asException());
+      return;
     }
     GetRecentTransactionsForAccountResponse response =
         GetRecentTransactionsForAccountResponse.newBuilder()
             .addAllTransactionEntry(transactionEntries)
             .build();
     responseObserver.onNext(response);
+    responseObserver.onCompleted();
   }
 
   private static AccountType toStorageAccountType(CreateAccountRequest.Type apiAccountType) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -197,12 +197,16 @@ public void getRecentTransactionsForAccount(
     } catch (SpannerDaoException e) {
       responseObserver.onError(Status.fromThrowable(e).asException());
       return;
+    } catch (IllegalArgumentException e) {
```
haseebajmal, line 200: nit: Maybe we can use a multi-catch here since the action we perform is the same for both ?

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -54,13 +54,15 @@
 
 @Category(IntegrationTest.class)
 public class FinAppIT {
-
-  private SpannerDaoInterface spannerDao;
-  private DatabaseClient databaseClient;
   private static Database db;
+  private DatabaseClient databaseClient;
+  private FinAppService finAppService;
```
haseebajmal, line 59: nit: Is this variable used anywhere ? We could perhaps just rename the blockingstub below to finAppService.

server/src/test/java/com/google/finapp/FinAppIT.java
```diff
@@ -156,41 +177,61 @@ public void createAccount_createsValidAccount() throws Exception {
     }
   }
 
-  @Test
-  public void createCustomerRole_createsValidCustomerRole() throws Exception {
-    ByteArray accountId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
-    ByteArray customerId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
-    ByteArray roleId = UuidConverter.getBytesFromUuid(UUID.randomUUID());
-    String roleName = "role name";
+  protected void addTestAccountRow(
```
haseebajmal, line 180: nit: private ?

haseebajmal (approved):

PR #61, opened 2021-09-28
Add ci-emulator workflow

voulg:
Add ci-emulator workflow


.github/workflows/ci-emulator.yaml
```diff
@@ -0,0 +1,42 @@
+on:
+  push:
+    branches:
+    - master
```
c24t, line 4:
```suggestion
    - main
```

.github/workflows/ci-emulator.yaml
```diff
@@ -0,0 +1,42 @@
...
+        image: gcr.io/cloud-spanner-emulator/emulator:latest
+        ports:
+          - 9010:9010
+          - 9020:9020
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: stCarolas/setup-maven@v4
+      with:
+        maven-version: 3.8.1
```
c24t, line 22: If we don't need a specific maven version we might want to avoid using third party actions: https://julienrenaux.fr/2019/12/20/github-actions-security-risk/. Not that we're passing in any secrets here though.

.github/workflows/ci-emulator.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - run: java -version
+    - run: gcloud config set auth/disable_credentials true
+    - run: gcloud config set project test-project
+    - run: gcloud config set api_endpoint_overrides/spanner http://localhost:9020/
+    - run: gcloud spanner instances create test-instance --config=emulator-config --nodes=1 --description="TestInstance"
+    - run: mvn clean compile
```
c24t, line 31: Do we need this? Looks like `compile` runs in the next step anyway. Do you expect `clean` to do anything on first run?

c24t (approved):
Only one blocking comment (`master` -> `main`), otherwise LGTM.

I put a few suggestions in https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/62 since I had to push these changes to a branch to test them anyway, up to you whether to include them in this branch or review/merge separately.

.github/workflows/ci-emulator.yaml
voulg: Merges your suggestion. I just copied whatever jdbc testing did, I'd assume that setting a maven version would make the test more stable, but perhaps let's keep things simple until we need the complexity.

Issue #46, opened 2021-08-23
Default timestamps for GetRecentTransactionsForAccount


biancamacias:
## Issue

Currently, the rpc GetRecentTransactionsForAccount does not have default timestamps. If only the `begin_timestamp` attribute is given, the default `end_timestamp` could be `Timestamp.MAX_VALUE` and vice versa.

c24t:
See https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/37#discussion_r695250963.

Issue #34, opened 2021-08-10
Consider creating a custom exception for invalid inputs


ngcgarcia:
## Issues

Currently, we throw `IllegalArgumentException` (an unchecked exception) for certain invalid inputs (i.e. negative amounts in `moveAccountBalance`). We catch these exceptions in `FinAppService` in order to provide useful feedback to users, but we could also be catching exceptions unrelated to the specific ones we're looking for (the ones we throw). We also expect these exceptions in our tests (namely, `moveAccountBalance`, which is being added soon) even though they are not in the signature. 

A possible fix for this is creating a custom checked exception. This would mean that we could include it in our method signatures to then test and that we could specifically catch this exception.

voulg:
Fixed by https://github.com/GoogleCloudPlatform/cloud-spanner-samples/pull/63

PR #63, opened 2021-10-01
More consistent exception handling. Default end_timestamp for getRTFA.

### LINKED ISSUE #33: Handle BigDecimal exceptions for invalid inputs
## Issue

We represent monetary amounts as strings in our proto and then convert these strings to `BigDecimal` to process in Java. In the case that an invalid string is given as input in [the constructor](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#BigDecimal-java.lang.String-), a `NumberFormatException` is thrown. We handle this for `createAccount` by catching the exception in `FinAppService`. This problem will apply to future PRs that require an amount as input. 

We should figure out a better way to handle this exception and provide useful output to users.

## Next steps

1. Consider different solutions for handling this exception, including possibly adding it to signatures
2. Change current methods that use this `BigDecimal` constructor (`createAccount`, `moveAccountBalance`, and more)
3. Ensure that future methods that will run into this problem will handle the exception similarly

voulg:
Fixes #33, #46 (and partially #34)

I also noticed we do not have very good coverage for the exceptions we throw, I plan to do this with a separate PR.

### DIFF
--- a/server/src/main/java/com/google/finapp/FinAppService.java
+++ b/server/src/main/java/com/google/finapp/FinAppService.java
@@ -59,24 +59,6 @@ final class FinAppService extends FinAppGrpc.FinAppImplBase {
     responseObserver.onCompleted();
   }
 
-  private BigDecimal getNonNegativeBigDecimal(String value) throws StatusException {
-    BigDecimal value_decimal = null;
-    try {
-      value_decimal = new BigDecimal(value);
-    } catch (NumberFormatException e) {
-      throw Status.INVALID_ARGUMENT
-          .withDescription(String.format("Invalid numeric value: %s", value))
-          .asException();
-    }
-    if (value_decimal.signum() == -1) {
-      throw Status.INVALID_ARGUMENT
-          .withDescription(
-              String.format("Expected positive numeric value, found: %s instead", value))
-          .asException();
-    }
-    return value_decimal;
-  }
-
   @Override
   public void createAccount(
       CreateAccountRequest account, StreamObserver<CreateAccountResponse> responseObserver) {
@@ -223,5 +205,23 @@ final class FinAppService extends FinAppGrpc.FinAppImplBase {
       default:
         return AccountStatus.UNSPECIFIED_ACCOUNT_STATUS;
     }
+  }
+
+  private BigDecimal getNonNegativeBigDecimal(String value) throws StatusException {
+    BigDecimal valueDecimal = null;
+    try {
+      valueDecimal = new BigDecimal(value);
+    } catch (NumberFormatException e) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription(String.format("Invalid numeric value: %s", value))
+          .asException();
+    }
+    if (valueDecimal.signum() == -1) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription(
+              String.format("Expected positive numeric value, found: %s instead", value))
+          .asException();
+    }
+    return valueDecimal;
   }
 }
--- a/server/src/main/java/com/google/finapp/FinAppService.java
+++ b/server/src/main/java/com/google/finapp/FinAppService.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import com.google.protobuf.ByteString;
 import io.grpc.Status;
+import io.grpc.StatusException;
 import io.grpc.stub.StreamObserver;
 import java.math.BigDecimal;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -58,28 +59,38 @@ public void createCustomer(
     responseObserver.onCompleted();
   }
 
+  private BigDecimal getNonNegativeBigDecimal(String value) throws StatusException {
```
haseebajmal, line 62: nit: I'd suggest moving this private method towards the end of the file so that it's consistent with the current ordering of the members.

server/src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -191,10 +176,21 @@ public void getRecentTransactionsForAccount(
     ByteArray accountId = ByteArray.copyFrom(request.getAccountId().toByteArray());
     Timestamp beginTimestamp = Timestamp.fromProto(request.getBeginTimestamp());
     Timestamp endTimestamp = Timestamp.fromProto(request.getEndTimestamp());
+    if (endTimestamp.equals(Timestamp.MIN_VALUE)) {
```
haseebajmal, line 179: nit: Can we instead do something like:
`Timestamp endTimestamp = request.hasEndTimestamp() ? Timestamp.fromProto(request.getEndTimestamp()) : Timestamp.MAX_VALUE`

It's not immediately obvious to me that Timestamp.fromProto(request.getEndTimestamp()) would result in MIN_VALUE when the field is empty.

haseebajmal (approved):

server/src/main/java/com/google/finapp/FinAppService.java
```diff
@@ -58,28 +59,38 @@ public void createCustomer(
     responseObserver.onCompleted();
   }
 
+  private BigDecimal getNonNegativeBigDecimal(String value) throws StatusException {
+    BigDecimal value_decimal = null;
```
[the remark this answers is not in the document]
voulg: Done

server/src/main/java/com/google/finapp/FinAppService.java
voulg: This is a good point. However, I was thinking in proto3 that does not have optional fields and the hasFieldX() API. Given that we're moving to proto3 API we'd eventually need to write this in this way.

server/src/main/java/com/google/finapp/FinAppService.java
voulg: done

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -114,13 +110,8 @@ public void createCustomerRole(
   @Override
   public ImmutableMap<ByteArray, BigDecimal> moveAccountBalance(
       ByteArray fromAccountId, ByteArray toAccountId, BigDecimal amount)
```
c24t, line 112: I see you catch negative `amount` in the caller now, but this is still a public method. Should negative `amount`s be legal here?

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -155,22 +148,19 @@ public void createCustomerRole(
...
   }
 
   @Override
   public BigDecimal createTransactionForAccount(
-      ByteArray accountId, BigDecimal amount, boolean isCredit) throws SpannerDaoException {
-    if (amount.signum() == -1) {
-      throw new IllegalArgumentException(
-          String.format("Amount transferred cannot be negative. amount: %s", amount.toString()));
-    }
+      ByteArray accountId, BigDecimal amount, boolean isCredit)
```
c24t, line 162: Same question about `amount` as above.

c24t (approved):
LGTM, only minor comments.

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -225,20 +220,13 @@ public BigDecimal createTransactionForAccount(
...
-          String.format("Invalid timestamp range. %s is before %s.", endTimestamp, beginTimestamp));
-    }
     Statement statement =
         Statement.newBuilder(
                 "SELECT * "
                     + "FROM TransactionHistory "
                     + "WHERE AccountId = @accountId AND "
-                    + "EventTimestamp BETWEEN @beginTimestamp AND @endTimestamp "
+                    + "EventTimestamp >= @beginTimestamp AND "
+                    + "EventTimestamp < @endTimestamp "
```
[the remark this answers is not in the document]
voulg: Correct, this will become important in the pagination PR coming soon.

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -199,10 +192,12 @@ public BigDecimal createTransactionForAccount(
...
 
                     if (newBalance.signum() == -1) {
-                      throw new IllegalArgumentException(
-                          String.format(
-                              "Account balance cannot be negative. original account balance: %s, amount to be removed: %s",
-                              oldBalance.toString(), amount.toString()));
+                      throw Status.INVALID_ARGUMENT
+                          .withDescription(
+                              String.format(
+                                  "Account balance cannot be negative. original account balance: %s, amount to be removed: %s",
```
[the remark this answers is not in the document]
voulg: Thanks I'll fix it in a follow-up PR

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
voulg: This is a non-public class, so it is ok-ish to assume things about its input. This is also documented in the interface were we mention that the amount is expected to be non-negative. I do see the concern, but I think in this case it's an internal implementation class and there's documentation in the interface so it's probably ok.

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
voulg: Resolving this and we can continue the conversation above.

PR #64, opened 2021-10-01
Use proto3

voulg:
Use proto3

### DIFF
--- a/server/src/main/proto/service.proto
+++ b/server/src/main/proto/service.proto
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-syntax = "proto2";
+syntax = "proto3";
 
 package com.google.finapp;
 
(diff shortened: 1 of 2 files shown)


server/src/main/proto/service.proto
```diff
@@ -42,85 +42,85 @@ service FinApp {
...
+  string new_balance = 1; // Digits[.[Digits]] or [Digits].Digits
 }
 
 message TransactionEntry {
-  optional bytes account_id = 1;
-  optional .google.protobuf.Timestamp event_timestamp = 2;
-  optional bool is_credit = 3;
-  optional string amount = 4;
+  bytes account_id = 1;
+  .google.protobuf.Timestamp event_timestamp = 2;
```
c24t, line 112: These can just be `Timestamp`s now right? https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp

c24t (approved):
LGTM after addressing the timestamp type.

server/src/main/proto/service.proto
voulg: I do not think so. The code you link is java not proto and the page says that these definitions are in the package google.protobuf . I also tried it and did not work... Let me know if there are reasons you expect it to work without the package name.

c24t: Nevermind, you're right.

PR #65, opened 2021-10-01
Voulg rtfa pagination

### LINKED ISSUE #49: Pagination for GetRecentTransactionsForAccount rpc
## Issue
Currently, the rpc needs 2 timestamps (begin and end) to query results.

## Next Steps
Come up with a way to query results in a `LIMIT BY` way. Beware that Spanner can perform poorly executing certain SQL statements.

voulg:
Fixes #49, also added more testing and fixed a few bugs.


c24t (approved):
LGTM, looks like a good change with or without pagination in the API -- which I assume we'll add in another PR.

voulg:
Actually this is enough to implement pagination! The way you'd do it is:
1. The first request does not set begin / end timestamp and sets the page limit as max_entry_count
For each subsequent page:
2. Re-run with end_timestamp = timestamp of oldest result you got so far and set the same page limit

Does this make sense?

PR #72, opened 2021-10-12
Add a bash script to aid developers run emulator/server/workload

haseebajmal:
Running all the commands (to bring up emulator, configure it, compile and bring up server/workload) from the README is cumbersome and error prone. This script IMO should help reduce some of the toil involved in development.

Usage:
```
$ bash run.sh emulator
$ bash run.sh server java \
        --spanner_project_id=test-project --spanner_instance_id=test-instance \
        --spanner_database_id=test-database
$ bash run.sh server jdbc \
        --spanner_project_id=test-project --spanner_instance_id=test-instance \
        --spanner_database_id=test-database
$ bash run.sh workload \
        --address-name localhost --port 8080 --thread-count 200
```

Updated the README to use this script.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -0,0 +1,110 @@
...
+    --config=emulator-config --description="Test Instance" --nodes=1
+  gcloud config set spanner/instance test-instance
+  gcloud spanner databases create test-database \
+    --ddl-file server/src/main/java/com/google/finapp/schema.sdl
+  export SPANNER_EMULATOR_HOST="localhost:9010"
+}
+
+run_server_java() {
+  mvn clean compile assembly:single -pl org.example:server
+  java -jar server/target/server-1.0-SNAPSHOT-jar-with-dependencies.jar $@
```
c24t, line 62: Any way to get the file name from the last command?

haseebajmal: Hm not quite sure. I think this can be replaced with `mvn exec java className` to make it slightly less brittle. I didn't want to deviate much from the existing instructions and just wrap them up in the script for convenience. We can perhaps revisit this if  this ends up causing maintenance burden.

PR #73, opened 2021-10-13
Python implementation of finapp server (CreateCustomer)

haseebajmal:
Implements CreateCustomer RPC in python.

- The code lives in `server/src/main/python`. It is organized into files similar to java.
- Uses [pip](https://pip.pypa.io/en/latest/user_guide/) to install dependencies. The dependencies are listed in `requirements.txt`.
- The server can be run in a virtual environment using the following commands:

  ```
  python -m venv env
  source env/bin/activate
  pip install -r server/src/main/python/requirements.txt
  export SPANNER_EMULATOR_HOST="localhost:9010"
  python server/src/main/python/server_main.py
  ```
  These commands are packaged into the dev script (run.sh) for convenience and the server can be brought up using a single command: `bash run.sh server python`.

- The proto sources are generated and checked in alongside source code. This seems to be common practice in python - Example: https://github.com/grpc/grpc/tree/master/examples/python/helloworld. Unlike maven, I couldn't find a way to auto generate the sources for us and clean them up as required. If needed we could potentially write the setup script to auto-generate the sources similar to https://github.com/protocolbuffers/protobuf/blob/master/python/setup.py (this seems to be the recommendation per various discussions online). However, manually generating them whenever protos change seems reasonable for this project. The python sources for protos can be generated using the following commands:

  ```
  python_source=server/src/main/python
  proto_source=server/src/main/proto
  python -m grpc_tools.protoc -I $proto_source \
    --python_out=$python_source --grpc_python_out=$python_source \
    service.proto
  protoc --proto_path=$proto_source --python_out=$python_source \
    database.proto
  ```
  These commands are packaged into the dev script (run.sh) for convenience and the sources can alternately be regenerated using: `bash run.sh gen_py_proto_srcs`

- Code is formatted using [yapf](https://github.com/google/yapf) {style: google, indent_width: 2, column_limit: 80}

Testing: I believe we would want to re-run the suite of tests `FinAppIT.java` for a python server. However, our setup currently doesn't allow us to do this. I plan on adding this support in subsequent PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@ for client libraries to work.
         --spanner_project_id=test-project --spanner_instance_id=test-instance \
         --spanner_database_id=test-database
     ```
-
     b. JDBC
 
     ```shell


README.md
```diff
@@ -13,17 +13,39 @@ for client libraries to work.
...
+    ```shell
     $ bash run.sh server java \
         --spanner_project_id=test-project --spanner_instance_id=test-instance \
         --spanner_database_id=test-database
     ```
-> To run the application using the JDBC implementation, in the command above,
-substitute `java` with `jdbc`.
 
-3. Call RPCs using grpc_cli.
+    b. Python
```
voulg, line 33: c. ?

run.sh
```diff
@@ -68,10 +80,19 @@ run_server_jdbc() {
     --spanner_use_jdbc $@
 }
 
+run_server_python() {
```
voulg, line 83: High-level question:
I tried to run the script in mac and had at least two issues:
- `declare -A` the bash version I had didn't know the -A option
- `fuser -K` the fuser in mac didn't know the -K option
I suspect that running in windows would be even more challenging.
The script works flawlessly in linux though.

Any ideas on how to improve compatibility? Some ideas:
a. Use python for scripting
b. Keep the script (for linux users), but also keep the more low-level instructions in case the script wouldn't work in the environment

This doesn't have to be done as part of this PR, but it's a good discussion to start.

server/src/main/python/finapp_service.py
```diff
@@ -0,0 +1,45 @@
...
+import grpc
+from google.api_core.exceptions import GoogleAPICallError
+from google.cloud import spanner
+
+import service_pb2
+import service_pb2_grpc
+from spanner_dao import SpannerDao
+
+
+class FinAppService(service_pb2_grpc.FinAppServicer):
```
voulg, line 26: Optional: Should we add more comments? (classes / methods)

run.sh
```diff
@@ -57,6 +59,16 @@ run_and_configure_emulator() {
   export SPANNER_EMULATOR_HOST="localhost:9010"
 }
 
+generate_py_proto_srcs() {
+  local python_source=server/src/main/python
+  local proto_source=server/src/main/proto
+  python -m grpc_tools.protoc -I $proto_source \
+    --python_out=$python_source --grpc_python_out=$python_source \
+    service.proto
+  protoc --proto_path=$proto_source --python_out=$python_source \
```
voulg, line 68: Where does `protoc` come from? I pip installed all the requirements but I do not have it in my linux environment.

server/src/main/python/spanner_dao.py
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import base64
+
+from google.cloud.spanner_v1 import Client
+
+
+class SpannerDao:
```
voulg, line 20: Optional: Comments?

server/src/main/python/spanner_dao.py
```diff
@@ -0,0 +1,32 @@
...
+
+  def __init__(self, client: Client, instance_id, database_id) -> None:
+    self._client = client
+    self._database = self._client.instance(instance_id).database(database_id)
+
+  def CreateCustomer(self, customer_id, name, address):
+    with self._database.batch() as batch:
+      batch.insert(
```
voulg, line 28: Why do we use batch API for this insert? I'd expect using transactions here.

PR #74, opened 2021-11-02
Use AccountStatus, Remove AccountType and clean-ups

### LINKED ISSUE #47: Define behavior for account statuses and types
## Issue

As the application is defined now, there is no difference in behavior between the unspecified, active, and frozen account statuses, along with the unspecified, checking, and saving account types.

## Next steps

1. Define the expected behavior for these different statuses/types for the different RPCs
2. If no difference in behavior is expected, consider the role and necessity these labels have in the app

voulg:
Fixes #47

- Remove AccountType as it was unused
- Use AccountStatus, in particular do not allow transfers for non-active status
- Refactoring to extract account information with a helper. This caught a bug where java implementation would throw when moving data between non-existing accounts
- Added tests for the bug and for frozen accounts
- Minor clean-ups


server/src/main/java/com/google/finapp/SpannerDaoImpl.java
```diff
@@ -251,20 +234,49 @@ public BigDecimal createTransactionForAccount(
...
+        throw Status.INVALID_ARGUMENT
+            .withDescription(String.format("Account not found: %s", accountId.toString()))
+            .asException();
+      } else if (accountData.get(accountId).status != AccountStatus.ACTIVE) {
+        throw Status.INVALID_ARGUMENT
+            .withDescription(
+                String.format(
+                    "Non-active accounts are not eligible for transfers: %s", accountId.toString()))
+            .asException();
+      }
```
c24t, line 277: This is a bit surprising since I'd assume `moveAccountBalance` is the part that cares whether the account is active.

c24t (approved):
Changes and tests LGTM, `AccountData` is a nice change too.

FWIW I don't see much more reason to keep `accountStatus` than `accountType` but maybe that will change soon.

voulg:
> FWIW I don't see much more reason to keep `accountStatus` than `accountType` but maybe that will change soon.

Well, now AccountStatus is used in the business logic, so it makes the example slightly more interesting. You could imagine Ops / Abuse automation marking accounts as frozen. Also due to the slight additional complexity I had to introduce the AccountData stuff... I can remove it if you prefer, but I think overall it adds to the example.

server/src/main/java/com/google/finapp/SpannerDaoImpl.java
voulg: I do not understand the comment. `readAccountDataForTransfer` is used for both `moveAccountBalance` and `createTransactionForAccount`. Given this method returns account data `ForTransfer` it does make sense to move eligibility checks here no?

c24t: I just mean that `readAccountDataForTransfer` could reasonably return data for non-active accounts based on the name, not a blocking comment.

voulg: You are right, it could be either. Actually, if the method was `readAccountData` it *should not* do the check. I explicitly went with the `ForTransfer` naming so I can move validations inside the helper to avoid duplication.

PR #77, opened 2021-11-11
Disallow balance move to same account

c24t:
Bugfix for `moveAccountBalance`, throw if `fromAccount` and `toAccount` are the same.


c24t:
Although we shouldn't be triggering this from the workload: https://github.com/GoogleCloudPlatform/cloud-spanner-samples/blob/main/workload/src/main/java/com/google/finapp/WorkloadClient.java#L62-L72.

voulg (commented):
It seems we tracked the bug, re-ping once fixed!

voulg (approved):
Thanks!

PR #80, opened 2022-01-18
Use docker to start emulator

zoercai:
Use the alternative docker way to start the emulator, making this compatible with the Kokoro VMs used for benchmarking tests.
https://cloud.google.com/spanner/docs/emulator#docker


c24t (approved):
LGTM, but you might consider using `--detach` instead of `&` with `docker run`.

PR #85, opened 2022-07-07
Use metaInf-services to merge files

voulg:
Files under resources/META-INF/services define service providers the application depends on, for more details see:
- https://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html

Our dependencies also define these files and in some cases they can overwrite our definitions. This change merges service providers from all dependencies with ours to ensure we have all needed dependencies. For more details see:
- https://maven.apache.org/plugins/maven-assembly-plugin/examples/single/using-container-descriptor-handlers.html


dtest (approved):
LGTM previous issues can't be repeated with this fix

PR #87, opened 2022-09-09
Add empty CloudSpannerSampleApp in SDL

c24t:
Add stub `CloudSpannerSampleApp` tables to let the Cloud Spanner UI know that this is a sample app.

### DIFF
--- a/finance/server/src/main/java/com/google/finapp/schema.sdl
+++ b/finance/server/src/main/java/com/google/finapp/schema.sdl
@@ -44,4 +44,8 @@ CREATE TABLE CustomerRole (
 ) PRIMARY KEY (CustomerId, RoleId),
   INTERLEAVE IN PARENT Customer ON DELETE CASCADE;
 
+CREATE TABLE CloudSpannerSampleApp (
+  id INT64,
+) PRIMARY KEY (id);
+
 CREATE INDEX CustomerRoleByAccount ON CustomerRole(AccountId, CustomerId)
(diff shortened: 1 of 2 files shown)


finance/server/src/main/java/com/google/finapp/schema.sdl
```diff
@@ -44,4 +44,8 @@ CREATE TABLE CustomerRole (
 ) PRIMARY KEY (CustomerId, RoleId),
   INTERLEAVE IN PARENT Customer ON DELETE CASCADE;
 
+CREATE TABLE CloudSpannerSampleApp (
+  id INT64,
```
voulg, line 48: Usually PRIMARY KEYs are NOT NULL.

Also consider removing trailing comma.

finance/server/src/main/java/com/google/finapp/schema_pg.sdl
```diff
@@ -48,4 +48,8 @@ CREATE TABLE CustomerRole (
     REFERENCES Account(AccountId)
 ) INTERLEAVE IN PARENT Customer ON DELETE CASCADE;
 
+CREATE TABLE CloudSpannerSampleApp (
+  id INT64,
+) PRIMARY KEY (id);
+
```
voulg, line 54: This should be postgres DDL, see above:
- INT64 -> BIGINT
- PRIMARY KEY definition should be inside the '(' ')'

voulg (commented):
Can you also test the ddl? I'm quite sure the PG was not valid

voulg (approved):

PR #93, opened 2024-05-21
Clean up README for Spanner finance app

varuncnaik:
* Changes "Cloud Spanner" to "Spanner".
* Adds required dependencies.
* Against a production Spanner instance, explains requirements for the project, instance, database, database DDL, and authentication.
* Adds syntax highlighting for multi-line shell snippets.
* Fixes quoting for grpc_cli protobufs. The protobuf argument in the shell should be enclosed in single quotes, so that the shell does not process special characters (e.g. `$\!`), especially when passing a value of type `bytes`.
* Clarifies when `SPANNER_EMULATOR_HOST` should be exported.
* Clarifies when a separate terminal window should be used.

### DIFF
--- a/finance/README.md
+++ b/finance/README.md
@@ -137,8 +137,8 @@ has the schema in
    > NOTE: To run the application using the JDBC implementation, in the command above,
 substitute `java` with `jdbc`.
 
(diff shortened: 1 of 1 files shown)


dtest (approved):
LGTM.

PR #99, opened 2025-07-28
manuelhurtado: LoR demo

mahurtado:
LoR Spanner Graph demo


Lsubatin (commented):
Thanks! waiting on headers and I'll merge. Added a single comment on the target for the backend in Cloud Run

Lsubatin:
Also, could we please add a file called CODEOWNERS with your github ID in it?

mahurtado:
Good catch! This file is only used for local testing. This is not used in
the demo deployed following the instructions.

